PR #52, opened 2016-02-17
Display error message if Avatar class was not defined

### LINKED ISSUE #46: Improve handling of code submission responses
Use http status codes to denote success and failure for code submission and change the returned data to be in some form of data structure (probably json) rather than checking whether `string.slice(0, prefix.length) == prefix`.

jishi9:
Fixes #46

### DIFF
--- a/aimmo-game/simulation/avatar/avatar_wrapper.py
+++ b/aimmo-game/simulation/avatar/avatar_wrapper.py
@@ -62,6 +62,8 @@ class AvatarWrapper(object):
             self.code = code
             try:
                 exec(code)
+                if Avatar is None:
+                    raise Exception('You need to define the class "Avatar"')
                 self.avatar = Avatar()
             except Exception as ex:
                 raise UserCodeException("Exception in user code", ex)


simulation/avatar/avatar_wrapper.py
```diff
@@ -60,6 +60,8 @@ def set_code(self, code):
         self.code = code
         try:
             exec(code)
+            if Avatar is None:
+                raise Exception('You need to define a class "Avatar"')
```
Spycho: Should this be a UserCodeException? And possibly have text of 'You need to define the class "Avatar"' or 'You need to define a class called "Avatar"'

Spycho: Ah, sorry, it will bubble to a UserCodeException anyway. Will that display sensibly to the user though, in terms of showing the right message?

mikebryant:
It's worth noting that when we move to running the user's code outside of the simulation thread, error handling will change significantly. We might need static validation when the user saves their code, or we might need to delay any errors until their code starts running in an avatar process.

#54 will stop this working, as saving of the code now doesn't result in it executing within the UI process.

jishi9:
> Will that display sensibly to the user though, in terms of showing the right message?

Yes:
image

>We might need static validation when the user saves their code, or we might need to delay any errors until their code starts running in an avatar process.

That's right. An idea that just occurred to me, so throwing it out there: validate user code by running it and passing it a fake but valid worldstate, just to make sure it doesn't crash immediately and does indeed return an action.
Just a thought.

mikebryant:
We're going to need a new approach for this, when we've fully realised the separated way of running the game & avatars, outside of the UI process.

(I'm imagining we have a more generic way of returning logging output from the users code, such that they can debug it - we can likely roll any parsing or runtime errors into that)

PR #63, opened 2016-05-04
WIP: Local Kubernetes Installation

mikebryant:


local-kubernetes/run
```diff
@@ -0,0 +1,46 @@
...
+set -e
+set -x
+
+# Make sure we're in the right directory
+cd $(cd -P -- "$(dirname -- "$0")" && pwd -P)
+
+export KUBERNETES_PROVIDER=vagrant
+export KUBERNETES_MEMORY=512
+
+if [ ! -d kubernetes ]; then
```
jishi9: I'm not sure this works? I tried `[[ -d rsync ]]` for example, and that returned false.
Maybe use `which kubernetes` instead?

mikebryant: That's testing whether the `kubernetes` directory exists.
If it doesn't exist, the get.k8s.io script will create it.

mikebryant:
Superseded by minikube

PR #71, opened 2016-05-11
Use socket.io for UI <-> Game comms. Fixes #56

### LINKED ISSUE #56: Change the way the UI receives game updates
The UI currently polls the frontend server constantly for updates.

Implement a solution where the *game simulation* server pushes updates instead - e.g. using web sockets (or some other better solution that you can think of!)

Possible server-side solutions:
1. client ==> proxy (on Kubernetes) ==> game simulation (on Kubernetes)
Here we only need one public IP address for the proxy, no matter how many game simulation servers we spawn.

2. client ==> game simulation (on Kubernetes)
We need a public IP address per game simulation. We need to investigate whether this would be a problem (e.g. can't be done dynamically, or is expensive).

mikebryant:
@jishi9 Try changing the sleep in `aimmo-game/simulation/turn_manager.py` to 0.1 or something.
Makes my browser a little slow, but it's cool to watch them move quickly.

Websockets ftw


jishi9:
Nice one! 😀 

Fun to watch for 5 seconds, and then it freezes and my laptop grinds to a halt... Chrome can't handle the update speed...

image

image

mikebryant:
This is where changing it to only get the data it needs, and only update the bits it needs might help, I guess.

I'm thinking it might be an idea to actually restructure the data stream going to the client to a pure series of events..

jishi9:
Running locally, and using a single web socket :)

image

PR #102, opened 2016-06-22
World map now expands in two directions (both away from 0,0)

observer61:
Can see this working if you change the map size (in aimmo-game service.py to start out at 3 by 3 and spinning up a couple of instances)


mikebryant:
Can these two commits be merged?

Spycho:
Isn't the rows variable also redundant?

Issue #68, opened 2016-05-04
Fog of War


mikebryant:
Don't send the entire map to the workers, have a configurable distance limit.
Open question: Manhattan distance, or rectangle?

OliverCartz:
Made a very simple rectangle fog.

To do anything more complicated might require drastic code changes to the way the map is stored and returned in various places

joshuablake:
Fixed by #108

PR #108, opened 2016-06-29
Fog of war

OliverCartz:
Branch related to #68


aimmo-game/simulation/world_map.py
```diff
@@ -12,6 +12,9 @@
 TARGET_NUM_PICKUPS_PER_AVATAR = 0.5
 PICKUP_SPAWN_CHANCE = 0.02
 
+NO_FOG_OF_WAR_DISTIANCE = 2
```
CelineBoudier: not distance?

aimmo-game/simulation/avatar/fog_of_war.py
```diff
@@ -7,4 +11,39 @@ def apply_fog_of_war(world_map, avatar_wrapper):
...
+def should_partially_fog(no_fog_distance, partial_fog_distance, x_dist, y_dist):
+    return x_dist > no_fog_distance or y_dist > no_fog_distance;
+
+def partially_fog_cell(cell):
+    partially_fogged_cell = copy.deepcopy(cell)
+    partially_fogged_cell.habitable = True
+    partially_fogged_cell.avatar = None
+    partially_fogged_cell.pickup = None
+    partially_fogged_cell.partially_fogged = True
+    return partially_fogged_cell;
```
Spycho: New line at EOF for PEP8 please

Spycho:
After discussion, we thought to go with missing attributes, rather than flagging what data is "known", or having an "unknown" constant.

joshuablake:
b227a2a has an alternate implementation which could allow removing `partially_fogged` from the API to address @mikebryant's concern. Also provides an easy way to add stuff to API with reduced duplication but more 'magic'.

aimmo-game-worker/simulation/world_map.py
```diff
@@ -67,7 +72,7 @@ def can_move_to(self, target_location):
             cell = self.get_cell(target_location)
         except KeyError:
             return False
-        return cell.habitable and not cell.avatar
+        return getattr(cell, 'habitable', False) and not getattr(cell, 'avatar', False)
```
joshuablake: This method is now a bit non-intuitive to me. "Unknown" (ie: a fogged cell) is returned as False. Proposed alternative behaviour (using None as an unknown flag):
* For unfogged `cell.habitable and not cell.avatar` (previous behaviour).
* For partially fogged: `None if cell.habitable else False`
* For fully fogged: `None`.

Maybe check if it's on the map for fogged cells (if we're giving avatars that info)?

aimmo-game-worker/simulation/world_map.py
```diff
@@ -51,7 +53,10 @@ def score_cells(self):
...
 
     def pickup_cells(self):
-        return (c for c in self.all_cells() if c.pickup)
+        return (c for c in self.all_cells() if getattr(self, 'pickup', False))
+
+    def partially_fogged_cells(self):
+        return (c for c in self.all_cells() if c.partially_fogged)
 
     def is_on_map(self, location):
         return location in self.cells
```
joshuablake: This now checks if the cell is in known/partially fogged (ie fully fogged cells give False). If we're not revealing map size then we should rename this method.

PR #109, opened 2016-06-29
Python implementation of run script

jishi9:


run.py
```diff
@@ -0,0 +1,57 @@
...
+            p.kill()
+        except:
+            pass
+
+
+def main():
+    server_args = []
+
+    run_command(['pip', 'install', '-e', '.'])
+    run_command(['./example_project/manage.py', 'migrate', '--noinput'])
```
Spycho: We'll likely need to put the python command in front of manage.py here, as on windows it doesn't know how to just execute the file. I'll test when I get a chance.

run.py
```diff
@@ -40,12 +40,12 @@ def main():
     server_args = []
 
     run_command(['pip', 'install', '-e', '.'])
-    run_command(['./example_project/manage.py', 'migrate', '--noinput'])
-    run_command(['./example_project/manage.py', 'collectstatic', '--noinput'])
+    run_command(['python', './example_project/manage.py', 'migrate', '--noinput'])
+    run_command(['python', './example_project/manage.py', 'collectstatic', '--noinput'])
```
CelineBoudier: can't we use os.path.join there instead of  / ?

run.py
jishi9: fixed

run.py
jishi9: done

CelineBoudier:
mayne squash the commits? Thanks :)

Issue #26, opened 2015-08-27
Make "power-ups" randomly appear and disappear


Spycho:
Similar to how scoring cells should in #25

- [x] Probably start with a "health" pickup
- [x] Then an attack damage power up designed to last a number of turns
- [x] Then an invulnerability power up designed to last a number of turns

PR #117, opened 2016-07-14
Add the "super power" pickups

joshuablake:
- Creates generic base classes for pickups and effects (a pickup applies an effect to an avatar).
- Moves the responsibility for taking a turn from TurnManager to AvatarWrapper (although this might want to be refactored elsewhere, especially the effect management part).
- Add an invulnerability and damage pickup.
- Adds basic frontend graphics for the above (blue and red circle respectively).

Completes backed for #26 although the graphics could probably be improved. Might also want to add effects to the serialisation of Avatars (although need to consider fog of war: should Avatars know effects that other Avatars have?).


aimmo-game/simulation/action.py
```diff
@@ -1,69 +1,74 @@
...
+        self._add_score_from_cell_if_needed(avatar, cell)
+        if cell.pickup is not None:
+            cell.pickup.apply(avatar)
+
+    def _add_score_from_cell_if_needed(self, avatar, cell):
+        if cell.generates_score:
+            avatar.score += 1
+
+    def __str__(self):
+        return self.__class__.__name__
```
jishi9: Nice :)

aimmo-game/simulation/avatar/avatar_wrapper.py
```diff
@@ -11,6 +19,36 @@ def __init__(self, initial_location, player_id, worker_url, avatar_appearance):
...
+        self.effects = set()
+        self.resistance = 0
+        self.attack_strength = 1
+
+    def take_turn(self, game_state, game_view):
+        action = self._get_action(game_view)
+        if action is not None:
+            action.apply(game_state, self)
+            LOGGER.debug("%s took %s" % (self.player_id, action))
+        effects_to_remove = set()
```
jishi9: `effects_to_remove = { effect for effect in self.effects if not effect.turn() }` might be more clear and more concise ?

aimmo-game/simulation/avatar/avatar_wrapper.py
```diff
@@ -11,6 +19,36 @@ def __init__(self, initial_location, player_id, worker_url, avatar_appearance):
...
+        for effect in self.effects:
+            if not effect.turn():
```
jishi9: it's not obvious in context what this means. Looking at the code I can infer that it means 'on_turn()', and returns a boolean to indicate whether the effect is now removed. Maybe it's clearer to have two separate methods for this? i.e. on_turn() and is_expired()

aimmo-game/simulation/avatar/avatar_wrapper.py
```diff
@@ -11,6 +19,36 @@ def __init__(self, initial_location, player_id, worker_url, avatar_appearance):
...
+                effects_to_remove.add(effect)
+        for effect in effects_to_remove:
+            effect.remove()
+
+    def _get_action(self, game_view):
+        try:
+            data = requests.post(self.worker_url, json=game_view).json()
```
jishi9: out of scope - but just bringing this up here:

does it really make sense to do POST http://my-worker/ ? Wouldn't POST http://my-worker/turn for example make more sense?

aimmo-game/simulation/avatar/avatar_wrapper.py
```diff
@@ -11,6 +19,36 @@ def __init__(self, initial_location, player_id, worker_url, avatar_appearance):
...
+        except ValueError as err:
+            LOGGER.info("Failed to get turn result: %s", err)
```
jishi9: I know it's the default, but maybe clearer to return None explicitly?

aimmo-game/simulation/avatar/avatar_wrapper.py
```diff
@@ -21,6 +59,11 @@ def die(self, respawn_location):
     def add_event(self, event):
         self.events.append(event)
 
+    def damage(self, amount):
+        applied_dmg = max(0, amount - self.resistance)
+        self.health -= applied_dmg
```
jishi9: is health ok to be negative?

jishi9: It seems like it is ok https://github.com/ocadotechnology/aimmo/pull/117/files#diff-d04a139bc2877a24168490f27b3091a2L54

aimmo-game/simulation/effects.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def turn(self):
+        self._time_remaining -= 1
+        return self._time_remaining != 0
+
+
+class InvulnerabilityEffect(_TimedEffect):
+    def __init__(self, *args):
+        super(InvulnerabilityEffect, self).__init__(*args)
+        self._avatar.resistance += 1000
```
jishi9: What happens if resistance is altered later on by something else? For example:
- start => resistance 5
- acquire InvulnerabilityEffect => resistance 1005
- Get hit by a 'lose resistance' object (this doesn't exist yet) => resistance 995
- expire InvulnerabilityEffect => resistance -5

aimmo-game/simulation/effects.py
```diff
@@ -0,0 +1,49 @@
...
+
+class DamageEffect(_TimedEffect):
+    def __init__(self, damage_boost, *args):
+        self._damage_boost = damage_boost
+        super(DamageEffect, self).__init__(*args)
+        self._avatar.attack_strength += self._damage_boost
+
+    def remove(self):
+        super(DamageEffect, self).remove()
+        self._avatar.attack_strength -= self._damage_boost
```
jishi9: potential similar problem as above

aimmo-game/simulation/pickups.py
```diff
@@ -0,0 +1,95 @@
...
+class InvulnerabilityPickup(_PickupEffect):
+    EFFECT = effects.InvulnerabilityEffect
+
+    def serialise(self):
+        return {
+                'type': 'invulnerability',
+        }
+
+
+class DamagePickup(_PickupEffect):
```
jishi9: Maybe DamageEffectPickup is clearer? DamagePickup sounds like a permanent change

aimmo-game/simulation/world_map.py
```diff
@@ -10,20 +11,7 @@
...
-        self.health_restored = health_restored
-
-    def __repr__(self):
-        return 'HealthPickup(health_restored={})'.format(self.health_restored)
-
-    def serialise(self):
-        return {
-            'health_restored': self.health_restored,
-        }
+PICKUP_SPAWN_CHANCE = 1
```
jishi9: Is this intentional?

aimmo-game/simulation/effects.py
joshuablake: The way you described the multiple effects is how I imagined it would be (ie by design). So you would effectively be invulnerable for the amount of time InvulnerabilityEffect applies for but afterwards would be more vulnerable for the remainder of 'lose resistance' effect.

jishi9: But presumably we want to prevent resistance from going negative right?

joshuablake: The base resistance is currently 0 so any 'lose resistance' effect is always going to cause negative resistance right?

jishi9: Yep. I was just questioning whether we want to have negative resistance (and what does that mean?) - otherwise we would need to (for e.g.) truncate it to zero.

If this is intentional then that's fine too :)

joshuablake:
(Above comments fixed, for some reason Github isn't recognising that the diffs are outdated)

aimmo-game/service.py
```diff
@@ -4,9 +4,8 @@
 import sys
 
 # If we monkey patch during testing then Django fails to create a DB enironment
```
Spycho: This will now monkey patch always. Is the above comment now invalid, or do we still need to avoid monkey patching during testing?

joshuablake: Comment is now invalid (tests are run separately to the Django tests now so there are no issues).

duncanrussell-ocado:
superseded by #158

PR #121, opened 2016-07-16
Django commands to generate players + delete generated players

jishi9:
This should be invoked via manage.py

@mikebryant I couldn't get this to work via manage.py , any ideas? :)

I tried placing the management/ directory in various locations, and I read through [this](https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/) and [this](https://community.webfaction.com/questions/16734/django-managepy-commands-not-displaying).


joshuablake:
For the command to work it needs to be part of a Django app, not in the project directory. Ie: `players/management/commands/generate_players.py` (with relevant __init__.py files).

jishi9:
Yes that worked, thanks @joshuablake :)

players/management/commands/generate_players.py
```diff
@@ -0,0 +1,98 @@
...
+
+        for _ in xrange(num_users):
+            random_string = str(uuid.uuid4())[:8]
+            username = 'zombie-%s' % random_string
+            password = '123'
+            user = self.create_user(username, password)
+            self.post_code(user, code)
+
+    def create_user(self, username, password):
+        user = User.objects.create_superuser(username, 'user@generated.com', password)
```
jishi9: Note to self: Is superuser needed?

jishi9:
I added a command to cleanup the generated users also.

I think this is ready for merge now :)

players/management/commands/generate_players.py
```diff
@@ -0,0 +1,98 @@
...
+
+from django.core.management import BaseCommand
+from django.test import RequestFactory
+from importlib import import_module
+from django.conf import settings
+from django.contrib.auth.models import User
+from players.views import code as code_view
+
+
+_AVATAR_CODES_DIRECTORY = 'players/avatar_examples'
```
joshuablake: Should be relative to this file otherwise cannot run command unless in the root directory.

players/management/commands/generate_players.py
```diff
@@ -0,0 +1,98 @@
...
+        super(Command, self).__init__(*args, **kwargs)
+        self.engine = import_module(settings.SESSION_ENGINE)
+        self.request_factory = RequestFactory()
+
+    def add_arguments(self, parser):
+        parser.add_argument('--num-users', type=int, required=True,
+                            help='Number of users to create')
+        parser.add_argument('--avatar-code', choices=_AVATAR_CODES,
+                            action=LoadCodeAction, required=True,
+                            help='The code to use for the avatar.')
```
joshuablake: I don't like the naming of the options here: one is users, the other is avatar, the command is players. I can't remember which is which!
Also, do we want required named rather than positional requirements ( `generate_players 3 dummy_avatar.py` vs `generate_players --num-users=3 --avatar-code=dummy_avatar.py`)

players/management/commands/generate_players.py
```diff
@@ -0,0 +1,98 @@
...
+
+        for _ in xrange(num_users):
+            random_string = str(uuid.uuid4())[:8]
+            username = 'zombie-%s' % random_string
+            password = '123'
+            user = self.create_user(username, password)
+            self.post_code(user, code)
+
+    def create_user(self, username, password):
+        user = User.objects.create_user(username, 'user@generated.com', password)
```
joshuablake: generated.com is a valid URL. Probably not an issue if we're only using this in testing though?

README.md
```diff
@@ -56,6 +56,10 @@ Even with these basic mechanics, there is quite a lot of complexity in creating
     * sync the database
     * collect the static files
     * run the server
+* You can quickly create players as desired using the following command:
+`python example_project/manage.py generate_players --num-users 5 --avatar-code dumb_avatar.py`
```
joshuablake: Document that the password is 123?

players/management/commands/generate_players.py
mikebryant: example.com is the canonical one for use in test data & documentation etc.

players/management/commands/generate_players.py
mikebryant: Standard for unix style commands (of which manage.py should be one) is positional arguments are required, and optional arguments are.. well, optional.
From the argparse docs: `Note Required options are generally considered bad form because users expect options to be optional, and thus they should be avoided when possible.`

joshuablake:
@jishi9 are you OK to make these changes? Otherwise I'm happy to take this over.

jishi9:
@joshuablake sorry I missed your comments, yes I'm happy to make the changes

players/management/commands/generate_players.py
jishi9: Done.

I'm not happy with the way I've had to do it - but I couldn't find an alternative in the BaseCommand class. Suggestions welcome!

players/management/commands/generate_players.py
jishi9: Changed to positional arguments.

Technically the generate_xx script creates both Django users and avatars for those Django users.
The deleted_generated_xx script deletes the generated Django users (and hence their avatars).


So I guess I can avoid saying 'players' at least. But I'm not sure what the right thing to do is. Should I just change everything to refer to avatars, even though it's actually creating/deleting Django users too?

players/management/commands/generate_players.py
jishi9: Done

README.md
jishi9: Done

players/management/commands/generate_players.py
joshuablake: Could use:
```
from players import avatar_examples
os.path.dirname(avatar_examples.__file__)
```

players/management/commands/delete_generated_players.py
```diff
@@ -0,0 +1,14 @@
+from django.core.management import BaseCommand
+from django.contrib.auth.models import User
+
+class Command(BaseCommand):
+    # Show this when the user types help
+    help = "Delete generated users"
+
+    # A command must define handle()
+    def handle(self, *args, **options):
+        for user in User.objects.iterator():
```
joshuablake: Why not `User.objects.filter(username__startswith='zombie-'`?

jishi9: I did not know this was a thing! TIL

Done.

Issue #115, opened 2016-07-13
Authentication for game-worker comms


mikebryant:
We should pass the worker a secret through the kubernetes API, and then it can use that to authenticate itself to the game

CelineBoudier:
What's the status on this? :)

OlafSzmidt:
@CelineBoudier Seems to be fixed in #136, but was never merged after that next pool of commits :(

OlafSzmidt:
Still relevant.

CelineBoudier:
Relevant

Issue #21, opened 2015-08-07
Switch 'code' field back to Avatar model from Player model


Spycho:
To support versioning functionality e.g. undo/redo. Also, implement undo/redo.

PR #122, opened 2016-07-18
Switch to using Avatar models and add Game model.

joshuablake:
Can now have multiple Avatars per user (eg: different games).
Opens up the way for a changelog on an Avatar.
Initial work for #115, #79, and #21.

### DIFF
--- a/players/models.py
+++ b/players/models.py
@@ -1,7 +1,23 @@
+from base64 import urlsafe_b64encode
 from django.contrib.auth.models import User
 from django.db import models
+from os import urandom
 
 
-class Player(models.Model):
-    user = models.OneToOneField(User)
+def generate_auth_token():
+    return urlsafe_b64encode(urandom(16))
+
+
+class Game(models.Model):
+    name = models.CharField(max_length=100)
+    auth_token = models.CharField(max_length=24, default=generate_auth_token)
+
+
+class Avatar(models.Model):
+    owner = models.ForeignKey(User)
+    game = models.ForeignKey(Game)
     code = models.TextField()
+    auth_token = models.CharField(max_length=24, default=generate_auth_token)
+
+    class Meta:
+        unique_together = ('owner', 'game')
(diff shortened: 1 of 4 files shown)


players/migrations/0003_auto_20160715_0914.py
```diff
@@ -0,0 +1,64 @@
...
+from django.conf import settings
+import players.models
+
+
+def move_data(apps, schema_editor):
+    Player = apps.get_model("players", "Player")
+    Avatar = apps.get_model("players", "Avatar")
+    Game = apps.get_model("players", "Game")
+
+    main_game = Game(pk=1, name="main")
```
mikebryant: We do not want any data addition in our migrations. We have this mistake in rapid router, I'd like to avoid having the same issue here.

i.e. If someone starts with a blank project, without existing data, and runs the migrations as part of normal setup, it should not result in there being any `Game` objects.

We could do something conditional, e.g. if there are any avatars, make a default game for them, otherwise do nothing...

players/models.py
```diff
@@ -1,7 +1,20 @@
...
+
+
+class Game(models.Model):
+    name = models.CharField(max_length=100)
+    auth_token = models.CharField(max_length=24, default=generate_auth_token)
+
+
+class Avatar(models.Model):
+    owner = models.ForeignKey(User)
+    game = models.ForeignKey(Game)
```
mikebryant: Do we want a `unique_together` on owner & game? e.g. 1 avatar per game

I'd say if we wanted to allow multiple avatars for a given user in a game, we should make that a property of the game and then run the code multiple times

joshuablake: Seems reasonable. Possible issue if we wanted to allow one user to have multiple avatars that acted differently (ie different code) in some kind of team game but that's probably far enough away that it can be ignored currently.

mikebryant: I considered that, but I think it would be better if the user accounted for that in their code (say, if we told the code how many copies were running and which ID that instance was)

joshuablake:
Suggested changes implemented in #124

PR #126, opened 2016-07-27
Add tests

joshuablake:
Covers aimmo-game and players (so the majority of the logic).


joshuablake:
Whenever I add tests to a simulation/test directory for a second app (eg aimmo-game-worker/simulation/test) I can't get anything to run due to having two top-level simulation packages. Tried playing around with sys.path but so far unsuccessful.

joshuablake:
OK, solved it (kinda). Need to play around with __path__.

mikebryant:
All the path manipulation is quite messy, and prone to being fragile.

I don't think we should be going this way - these are separate pieces, not parts of the same python application - we shouldn't be testing them in one unit.
(And we should probably be thinking about splitting them out into different repositories, if we're having this sort of difficulty testing them)

They should really have separate test scripts, so we have `python setup.py test` for the django app, but the docker images should be being tested differently, as they aren't really python packages

joshuablake:
Don't see anyway to fix Coveralls without either using `coverage run setup.py test` or writing a custom test suite with coverage.py API. Although I'm not sure why it isn't picking up at least the Django tests.
Apart from that, think this is done now (unless decide to split into separate repos)?

Issue #118, opened 2016-07-14
Add a concurrent way of running turns


mikebryant:
Give current world state to all workers, get all actions, run some sort of conflict resolution

This should be configurable on a per-game basis (i.e. this new one or the current algorithm)

In theory the old algorithm will be a subset, as you could just call the new one with one action each time so there'd be no conflict resolution..

joshuablake:
Main conflict seems to be if two avatars want to move onto the same square. Options for resolution:

- Both fail (probably the simplest)
- Randomly choose one to succeed
- Some kind of heuristic based on the code (runtime of each player's code?)
- A fight of some kind (make both attack each other until one is dead?)

I like both failing as it strikes me as the fairest and adds an incentive to predict other user's behaviour. Although it could become frustrating on a crowded grid.

mikebryant:
Though it does get interesting to implement, as if you had:
`A B _ C`
if A & B move right to the empty square, and C moves left, B & C conflict, which then means if they both fail, A can't move either. It's not obvious to me how to resolve that in one pass quickly..

joshuablake:
You can do it by storing all the moves then backtracking when there's a conflict. [Proof of concept.](https://gist.github.com/joshuablake/8200ddb3d586933a6c87cedba47aaa94)

joshuablake:
Do we want to allow avatars to move pass each other (so `A B` into `B A`)? Seems weird they can move "through" each other but possibly expensive to check.

Spycho:
We also need to make sure that the conflict resolution aspect doesn't make the game prohibitively hard for users to understand and program for. We should keep this in mind when choosing a resolution strategy.

PhydeauxLeChien:
Resolved by #127.

PR #127, opened 2016-07-28
Concurrent turns

PhydeauxLeChien:
- `run_turn` now performs each worker request in a new thread
- actions are registered on the world map before they are applied
- once all actions are registered, actions are applied in this order:
    - wait actions
    - attack actions
    - move actions
- avatars may be attacked in both their current location and their new location
- if two or more avatars attempt to move to the same location, they all fail

### DIFF
--- a/aimmo-game/simulation/action.py
+++ b/aimmo-game/simulation/action.py
@@ -22,8 +22,8 @@ class Action(object):
         return self._target_location
 
     def register(self, world_map):
-        if world_map.is_on_map(self._target_location):
-            world_map.get_cell(self._target_location).actions.append(self)
+        if world_map.is_on_map(self.target_location):
+            world_map.get_cell(self.target_location).actions.append(self)
 
     def process(self, world_map):
         if self.is_legal(world_map):
@@ -59,43 +59,51 @@ class MoveAction(Action):
         super(MoveAction, self).__init__(avatar)
 
     def is_legal(self, world_map):
-        return world_map.can_move_to(self._target_location)
+        return world_map.can_move_to(self.target_location)
 
     def process(self, world_map):
-        if self.is_legal(world_map):
-            self.chain(world_map, self)
-        else:
-            self.reject()
+        self.chain(world_map, {self.avatar.location})
 
     def apply(self, world_map):
-        event = MovedEvent(self.avatar.location, self._target_location)
+        event = MovedEvent(self.avatar.location, self.target_location)
         self.avatar.add_event(event)
 
         world_map.get_cell(self.avatar.location).avatar = None
-        self.avatar.location = self._target_location
-        world_map.get_cell(self._target_location).avatar = self.avatar
+        self.avatar.location = self.target_location
+        world_map.get_cell(self.target_location).avatar = self.avatar
         self.avatar.clear_action()
 
-        new_cell = world_map.get_cell(self._target_location)
+        new_cell = world_map.get_cell(self.target_location)
         if new_cell.pickup:
             # TODO:  extract pickup logic into pickup when adding multiple types
             self.avatar.health = min(10, self.avatar.health + new_cell.pickup.health_restored)
             new_cell.pickup = None
 
-    def chain(self, world_map, first):
-        if self.is_legal(world_map):
-            avatar = world_map.get_cell(self.target_location).avatar
-            if avatar is None or (avatar.action is not first and
-                                  avatar.action.chain(world_map, first)):
-                self.apply(world_map)
-                return True
-        self.reject()
-        return False
+        return True
+
+    def chain(self, world_map, visited):
+        if not self.is_legal(world_map):
+            return self.reject()
+
+        # Detect cycles
+        if self.target_location in visited:
+            return self.reject()
+
+        next_cell = world_map.get_cell(self.target_location)
+        if not next_cell.is_occupied:
+            return self.apply(world_map)
+
+        next_action = next_cell.avatar.action
+        if next_action.chain(world_map, visited | {self.target_location}):
+            return self.apply(world_map)
+
+        return self.reject()
 
     def reject(self):
-        event = FailedMoveEvent(self.avatar.location, self._target_location)
+        event = FailedMoveEvent(self.avatar.location, self.target_location)
         self.avatar.add_event(event)
         self.avatar.clear_action()
+        return False
 
 
 class AttackAction(Action):
@@ -105,13 +113,13 @@ class AttackAction(Action):
         super(AttackAction, self).__init__(avatar)
 
     def is_legal(self, world_map):
-        return True if world_map.attackable_avatar(self._target_location) else False
+        return True if world_map.attackable_avatar(self.target_location) else False
 
     def apply(self, world_map):
-        attacked_avatar = world_map.attackable_avatar(self._target_location)
+        attacked_avatar = world_map.attackable_avatar(self.target_location)
         damage_dealt = 1
         self.avatar.add_event(PerformedAttackEvent(attacked_avatar,
-                                                   self._target_location,
+                                                   self.target_location,
                                                    damage_dealt))
         attacked_avatar.add_event(ReceivedAttackEvent(self.avatar,
                                                       damage_dealt))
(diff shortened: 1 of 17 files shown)


Spycho:
Is the behaviour of concurrent vs sequential determined based on an attribute of a game? Sorry, haven't had a chance to look through the code yet.

aimmo-game/simulation/turn_manager.py
```diff
@@ -20,19 +21,20 @@ def __init__(self):
         self._world_state = None
```
joshuablake: Maybe rename to game_state? Not clear if it's a WorldMap or a GameState.

aimmo-game/simulation/world_map.py
```diff
@@ -149,9 +157,25 @@ def get_random_spawn_location(self):
...
+        if not self.is_on_map(target_location):
+            return False
         cell = self.get_cell(target_location)
-        return cell.habitable and not cell.avatar
+
+        if cell.avatar:
+            return cell.avatar
+
+        if len(cell.actions) == 1:
+            return cell.actions[0].avatar
```
joshuablake: This causes an avatar to attack itself if this is the only action on target_location.

joshuablake:
I don't think your code handles [the case when backtracking is required](https://github.com/ocadotechnology/aimmo/issues/118#issuecomment-232668082). Or at least, the behaviour is dependant on the order in which the actions are placed into the queue.
@Spycho no, it's always concurrent.

PhydeauxLeChien:
@Spycho I will add the option to revert to sequential turns shortly.

PhydeauxLeChien:
@joshuablake good spot, that should be the list of move actions on that cell only.

PhydeauxLeChien:
@joshuablake you mean `A B _ C` where `A` and `B` both want to move right and `C` wants to move left? In this implementation they will all fail to move, as required. However, `A` will sometimes fail to move even if `B` succeeds in moving right (depending on their position in the queue, as you say), which we probably want to allow.

In order to implement the desired behaviour, I plan to apply the moves in multiple rounds - in each round remove the already successful moves and attempt to apply the unsuccessful moves again, terminating when there is no change in the list of unsuccessful moves.

aimmo-game/simulation/turn_manager.py
PhydeauxLeChien: an artifact from renaming `WorldStateManager` to the (imo) less confusing `GameStateManager`

aimmo-game/simulation/turn_manager.py
```diff
@@ -80,13 +84,18 @@ def run_concurrent_turn(self):
...
+            with game_state_provider as game_state:
+                if action.is_legal(game_state.world_map):
+                    action.apply(game_state.world_map)
+                else:
+                    action.reject()
+            cells_to_clear.add(action.target_location)
+
+        for cell in cells_to_clear:
+            with game_state_provider as game_state:
                 game_state.world_map.clear_cell_actions(cell)
```
joshuablake: Think we should call reject() on cleared actions to register the relevant events.

PhydeauxLeChien: all events cleared here will already either have been applied or rejected.

PhydeauxLeChien:
TODO: change the move_chain method to keep track of the set of visited cells, instead of just the first in the chain.

joshuablake:
Collecting here a list of cases we should test, excluding basic ones with no conflict resolution. Need t test each with a variety of ordering in chain.

Arrows represent requested movement direction, `x` is wait.

## Chains
### One direction, passes
`> > > > _`, expected: ` _ x x x x`.

### One direction, fails
`> > > x`, expect `x x x x`.

### Snaking
```
> > ▼
< < <
```
Expected:
```
_ _ x x
x x x x
```

## Two directions
`> > _ <`, expect `x x _ x`.

## Loops
### Complete

```
> ▼
^ < 
```

Expect no moves (or all move)

### Complete, with extra
Especially with the top left as both start and end of chain:

```
> > ▼
  ^ < 
```

should fail.

## Swapping
`> <`, expect both moves fail.

aimmo-game/simulation/turn_manager.py
```diff
@@ -1,81 +1,123 @@
...
-        try:
-            game_state = world_state_provider.lock_and_get_world()
-
-            for avatar in game_state.avatar_manager.active_avatars:
-                turn_state = game_state.get_state_for(avatar)
-                try:
-                    data = requests.post(avatar.worker_url, json=turn_state).json()
-                except ValueError as err:
-                    LOGGER.info("Failed to get turn result: %s", err)
+        if self.concurrent_turns:
```
joshuablake: To me this is a code smell that these should be separated into their own subclasses.

aimmo-game/simulation/turn_manager.py
```diff
@@ -1,81 +1,123 @@
...
+            self.run_concurrent_turn()
+        else:
+            self.run_sequential_turn()
+
+    def run_sequential_turn(self):
+        '''
+        Get and apply each avatar's action in turn.
+        '''
+        with state_provider as game_state:
+            avatars = game_state.avatar_manager.active_avatars
```
joshuablake: Do we want to release the lock here? We immediatly get it back inside [_register_action](https://github.com/ocadotechnology/aimmo/pull/127/files#diff-0df92163d570b5e36adea6e4166f2f9aR107) and also the loop. We could end up giving different states to different avatars.

Slightly off-topic: I'm not sure what other thread uses the state_provider lock? WorkerManager has its own lock. Could we have race conditions between TurnManager and WorkerManager (eg if an avatar is removed during a turn)?

aimmo-game/simulation/action.py
```diff
@@ -1,75 +1,154 @@
...
+        raise NotImplementedError('Abstract method')
+
+    def apply(self, world_map):
+        raise NotImplementedError('Abstract method')
+
+    def reject(self):
         raise NotImplementedError('Abstract method')
 
 
 class WaitAction(Action):
```
joshuablake: Needs a reject method?

aimmo-game/simulation/action.py
```diff
@@ -1,75 +1,154 @@
...
-    def apply(self, game_state, avatar):
-        _add_score_from_cell_if_needed(avatar, game_state)
+    def __init__(self, avatar):
+        super(WaitAction, self).__init__(avatar)
+
+    def is_legal(self, world_map):
+        return True
+
+    def apply(self, world_map):
+        self.avatar.clear_action()
```
joshuablake: WaitAction needs a reject method

aimmo-game/simulation/action.py
```diff
@@ -1,75 +1,154 @@
...
+        if next_action.chain(world_map, visited | {self.target_location}):
+            return self.apply(world_map)
+
+        return self.reject()
+
+    def reject(self):
+        event = FailedMoveEvent(self.avatar.location, self.target_location)
+        self.avatar.add_event(event)
+        self.avatar.clear_action()
+        return False
```
joshuablake: What does the return value signify?

aimmo-game/simulation/turn_manager.py
```diff
@@ -1,81 +1,123 @@
...
+
+        for avatar in avatars:
+            self._register_action(avatar)
+            with state_provider as game_state:
+                if avatar.action.is_legal(game_state.world_map):
```
joshuablake: As discussed, change to call process

aimmo-game/simulation/turn_manager.py
PhydeauxLeChien: yes, I was considering doing this

aimmo-game/simulation/action.py
PhydeauxLeChien: it should never be called since `is_legal` is defined as `true`, so I think the `NotImplementedError` is desirable if it were ever called

aimmo-game/simulation/action.py
PhydeauxLeChien: Filthy, filthy hack to avoid having to type `self.reject() ; return False` a bunch of times in `MoveAction.chain` below.

joshuablake:
Edge case we discussed earlier (arrows represent attack direction):
```
  B 
> ^ <
```

1. The most westerly avatar has its action applied first, causing the middle avatar to die.
2. The avatar that just died has its action applied, causing B to die (expected).
3. B re-spawns in the (now vacant) cell that the middle avatar was in.
4. The final action (from the easterly avatar) is processed, which damages B (unwanted).

This occurs when there is a valid target for an attack at the start of the turn and a valid target when it is processed but these targets are not the same avatar.

5e51d87 contains a unittest for this.

PR #129, opened 2016-08-04
Add script to run all tests

joshuablake:
* Script that runs `python setup.py test` for each app.
* Add shebang to run.py.

### DIFF
--- a/all_tests.py
+++ b/all_tests.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+import os
+import subprocess
+import sys
+
+
+def main():
+    BASE_DIR = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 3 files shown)


all_tests.py
```diff
@@ -0,0 +1,27 @@
...
+import subprocess
+import sys
+
+
+def main():
+    BASE_DIR = os.path.abspath(os.path.dirname(__file__))
+    APPS = ('', 'aimmo-game', 'aimmo-game-worker', 'aimmo-game-creator')
+    failed_apps = []
+    for app in APPS:
+        os.chdir(os.path.join(BASE_DIR, app))
```
jishi9: Subprocess.call takes a 'cwd' argument

joshuablake: What I get for reading the documentation too quickly. Thanks.

PR #130, opened 2016-08-04
Expand in 4-directions

joshuablake:
Map is now centred on origin and will expend in all four compass directions (always adds all four sides each expansion).

### DIFF
--- a/aimmo-game/service.py
+++ b/aimmo-game/service.py
@@ -33,13 +33,13 @@ def to_cell_type(cell):
     return 0
 
 
-def player_dict(avatar):
+def player_dict(avatar, min_x, min_y):
     # TODO: implement better colour functionality: will eventually fall off end of numbers
     colour = "#%06x" % (avatar.player_id * 4999)
     return {
         'id': avatar.player_id,
-        'x': avatar.location.x,
-        'y': avatar.location.y,
+        'x': avatar.location.x-min_x,
+        'y': avatar.location.y-min_y,
         'health': avatar.health,
         'score': avatar.score,
         'rotation': 0,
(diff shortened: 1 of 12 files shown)


aimmo-game/service.py
```diff
@@ -35,13 +35,13 @@ def to_cell_type(cell):
...
 
-def player_dict(avatar):
+def player_dict(avatar, min_x, min_y):
     # TODO: implement better colour functionality: will eventually fall off end of numbers
     colour = "#%06x" % (avatar.player_id * 4999)
     return {
         'id': avatar.player_id,
-        'x': avatar.location.x,
-        'y': avatar.location.y,
+        'x': avatar.location.x-min_x,
```
mikebryant: This sounds like it would be a significant change from the current player-facing api.

If you can't rely on the coordinates of avatars or bits of the map remaining constant from one call to the next it's going to make persisting world data in a fog situation really confusing.

(We've been imagining that if you're writing an avatar for a simulation with fog, you can persist the data to your local avatar each time, to gradually build up a picture of the whole world. If these coordinates can change it's going to make that confusing for students.)

joshuablake: This is only used for the UI API (used by the JS front-end). The location for the API exposed to avatar's is based off each class' serialise method.

mikebryant: Oh, I see. never mind then, my mistake

observer61:
Seems to be many changes being done to account for the transformation.

To prevent bugs coming in (from someone simply forgetting to do this for a new feature or making a slight miscalculation) it would be better if the transformation could be done when transferring data from the back-end to the front end.

So that all the calculations/logic it's all in one place.
(although I don't yet know where that could be)

aimmo-game/simulation/map_generator.py
```diff
@@ -85,31 +84,30 @@ def manhattan_distance_to_destination_cell(this_branch):
...
-    num_row_cells = width - 2
-    num_col_cells = height - 2
+def get_random_edge_index(map, rng=random):
+    num_row_cells = map.num_rows - 2
+    num_col_cells = map.num_cols - 2
     num_edge_cells = 2*num_row_cells + 2*num_col_cells
     random_cell = rng.randint(0, num_edge_cells-1)
-
-    if 0 <= random_cell < num_row_cells:
+    # import pdb; pdb.set_trace()
```
joshuablake: Whoops

aimmo-game/simulation/world_map.py
```diff
@@ -68,10 +68,23 @@ class WorldMap(object):
...
     def __init__(self, grid):
-        self.grid = grid
+        self._grid = grid
+
+    @classmethod
+    def generate_empty_map(cls, height, width):
+        start_x = -width / 2 + 1
+        end_x = width / 2 + 1
+        start_y = -height / 2 + 1
+        end_y = height / 2 + 1
```
observer61: I think it was mentioned these implicitly floor when you do (width / 2).
Maybe it is because I am unfamiliar with python , but explicit flooring would be preferable)

aimmo-game/simulation/world_map.py
```diff
@@ -85,22 +98,77 @@ def pickup_cells(self):
...
+        max_y = 0
+        while True:
+            try:
+                self._grid[Location(0, max_y)]
+            except KeyError:
+                max_y -= 1
+                break
+            else:
+                max_y += 1
+        return max_y
```
observer61: These methods of going till your out of the map and then re-adjusting by one seem a bit hacky.
I prefer your suggestion of maintaining the value of the variable when outer layers are added (andpossibly removed in the future)

aimmo-game/simulation/world_map.py
CelineBoudier: Well it's a normal Python 2.7 fact. However, the behaviour of / has changed with Python 3, where it will return a float. So explicit flooring might be a good idea if we plan forward compatibility with Python3 (which we might like to use at some point...)

joshuablake:
Reworked JS so that it can use the negative co-ordinates directly.

Tidied up some code brought up by review.

aimmo-game/simulation/world_map.py
```diff
@@ -89,10 +89,28 @@ class WorldMap(object):
...
 
     def __init__(self, grid):
-        self.grid = grid
+        self._grid = grid
+
+    @classmethod
+    def _min_max_from_dimensions(cls, height, width):
+        max_x = int(math.floor(width / 2))
+        min_x = -(width - max_x - 1)
+        max_y = int(math.floor(height / 2))
```
CelineBoudier: there are other /2 elsewhere in your PR if you wish to be consistent with the use of floor :)

Spycho:
Merging locally, fixing merge issues, fixing pep issues

Issue #113, opened 2016-07-13
Replace Ingress


mikebryant:
We need a local service doing the reverse proxy, and a single ingress just pointing to that service

Because:
* Ingress has a limit on backends
* Reconfiguring the ingress then means it takes a while before it notices the new game

mikebryant:
Note, instead we'd have a static ingress, pointing to our reverse proxy implementation

joshuablake:
The only suitable pre-built solution that I could find is [Ceryx](https://github.com/sourcelair/ceryx).

joshuablake:
Fixed by #147

PR #135, opened 2016-08-10
Add script to run minikube

joshuablake:
Todo:

- [x] Script to install minikube/kubectl locally
- [x] Create all Pods using local images.
- [x] Change API environment variables based on whether running locally or in the cloud. (How is this currently done when moving from staging to production? @mikebryant @Spycho )
- [x] Find a workaround for lack of ingress support (related: #113).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ db.sqlite3
 
 # static folder
 example_project/example_project/static/
+
+# Created during minikube testing
+test-bin/
--- a/README.md
+++ b/README.md
@@ -72,6 +72,27 @@ Even with these basic mechanics, there is quite a lot of complexity in creating 
 
 `python example_project/manage.py delete_generated_players`
 
+### Under Kubernetes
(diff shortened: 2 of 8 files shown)


joshuablake:
Seems to be mostly working but runs into resource limits with more than ~10 avatars.

joshuablake:
Edge case we discussed earlier (arrows represent attack direction):
```
  B 
> ^ <
```

1. The most westerly avatar has its action applied first, causing the middle avatar to die.
2. The avatar that just died has its action applied, causing B to die (expected).
3. B re-spawns in the (now vacant) cell that the middle avatar was in.
4. The final action (from the easterly avatar) is processed, which damages B (unwanted).

This occurs when there is a valid target for an attack at the start of the turn and a valid target when it is processed but these targets are not the same avatar.

5e51d87 contains a unittest for this.

joshuablake:
Appears to be working fine on Linux (Ubuntu). In theory should support OSX and Windows as well. Would be useful if someone could test (@Spycho ?)

Issue #114, opened 2016-07-13
Put SSL on Ingress


mikebryant:
So we don't need to keep loading unsafe scripts

CelineBoudier:
@adstastic @ChrisMukherjee Any news? :)

CelineBoudier:
Hmm well :/

PR #147, opened 2016-08-23
Replace Ingress with a reverse proxy

joshuablake:
* Run a nginx pod which does the same job ingress was doing but use the Kubernetes DNS instead of updating with aimmo-game-creator.
* Add this pod to a static ingress.

Config that will need updating:
* Add the new image to Docker Hub.
* Launch the new pod in the update script.
* Update the DNS alias for staging.aimmo.codeforlife.education to point to the new ingress IP

### DIFF
--- a/aimmo-reverse-proxy/nginx.conf
+++ b/aimmo-reverse-proxy/nginx.conf
@@ -1,0 +1,13 @@
+
+http {
+    resolver 10.0.0.10;
+    server {
+        listen 80;
+        location ~^/game/([a-zA-Z]+)/? {
+            proxy_pass "http://game-$1.default.svc.cluster.local/";
+        }
+    }
+}
+
+events {
+}
--- a/aimmo-game-creator/service.py
+++ b/aimmo-game-creator/service.py
@@ -3,7 +3,7 @@
 import logging
 from pykube import HTTPClient
 from pykube import KubeConfig
-from pykube import Ingress, ReplicationController, Service
+from pykube import ReplicationController, Service
 import time
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


aimmo-reverse-proxy/nginx.conf
```diff
@@ -0,0 +1,13 @@
+
+http {
+    resolver 10.0.0.10;
+    server {
+        listen 80;
+        location ~^/game/([a-zA-Z]+)/? {
+            proxy_pass "http://game-$1.default.svc.cluster.local/";
```
mikebryant: It's generally better not to hardcode the cluster dns name, or the namespace - the search path should be set up correctly to allow for this.

e.g.
```
proxy_pass "http://game-$1/";
```

mikebryant:
hub.docker.com config is set up

aimmo-reverse-proxy/nginx.conf
joshuablake: I tried this but cannot be used as nginx does not use resolv.conf for dynamic DNS lookups. (It won't use `gethostbyname` because it's blocking.)

The alternative is to try and do something with manually parsing resolv.conf in Dockerfile.

mikebryant: Oh :(

Fair enough then, I'd say leave as is

mikebryant:
LGTM

joshuablake:
Broken on staging: the main site is under HTTPS but ingress (I'm guessing that should be handling SSL not nginx) doesn't which means that you can't send XHR requests. Was this working before, if not assume it's due to #114.

mikebryant:
Yeh, this has always been broken. You can use the grey shield icon in the urlbar in chrome to override it temporarily for testing.

PR #150, opened 2016-08-24
Use deployments for workers.

joshuablake:
- Workers automatically restart if they die.
- Each worker has a service to abstract this possibility from the game.

Should probably change so that rather than deleting/recreating a worker on code change instead just kill the pod (the remade pod will then pull the changed code).


Spycho:
I'm not sure quite how to resolve the merge conflicts, as master now does some extra bits. @mikebryant might be able to figure it out? There are only three conflicts in one file.

OlafSzmidt:
# Closing this pull request to stale it. We will review this at a later stage and perhaps re-open it.

PR #153, opened 2016-08-31
Big refactoring of `GameState`

PhydeauxLeChien:
Lots of changes, with the overall goal of improving / simplifying the way in which actions, pickups, and other yet-to-be-invented effects interact with the game state.
- `GameState` now properly thread-safe with its own lock and context manager methods.
- New decorator `@alters_state` for methods on `GameState` to ensure that lock is acquired before making changes.
- `WorldMap`, `AvatarManager`, and `AvatarWrapper` now read and modified only through `GameState` API - more methods to be added (e.g. for pickup effects) but all should atomically move from one valid state to another.
- Intended actions for concurrent turns now stored in an `ActionRegistry` instead of multiple places within game state.


aimmo-game/simulation/location.py
```diff
@@ -1,10 +1,24 @@
...
+    def y(self):
+        return self._y
 
     def __add__(self, direction):
-        return Location(self.x + direction.x, self.y + direction.y)
+        try:
+            d_x = direction.x
+            d_y = direction.y
+        except AttributeError:
+            return self
```
joshuablake: I don't like the fact that adding a Location to something nonsensical (eg an integer) will fail silently.

aimmo-game/simulation/action.py
```diff
@@ -1,145 +1,152 @@
...
 class Action(object):
-    def __init__(self, avatar):
-        self._avatar = avatar
-        try:
-            self._target_location = self._avatar.location + self.direction
-        except AttributeError:
-            self._target_location = self._avatar.location
+    def __init__(self, avatar_id, origin, direction=None):
+        self._avatar_id = avatar_id
+        self._origin = origin
```
joshuablake: origin is a bit of an ambiguous name - I expect it to refer to (0, 0).

aimmo-game/simulation/turn_manager.py
```diff
@@ -79,42 +39,94 @@ def run_turn(self):
...
+        return self._actions_by_target.setdefault(target, [])
+
+    def by_target(self, target):
+        return (action for action in self._by_target(target))
+
+    def moves_to(self, target):
+        return (a for a in self.by_target(target) if isinstance(a, MoveAction))
+
+    def num_moves_to(self, target):
+        return sum(1 for move in self.moves_to(target))
```
joshuablake: This is really unclear, how about `len(list(self.moves_to(target)))`? Higher memory overhead althouh [might be faster](http://stackoverflow.com/questions/393053/length-of-generator-output#comment48645858_7223557).

aimmo-game/simulation/location.py
PhydeauxLeChien: Yeah, this was only for handling `direction == None`. I'll do it a better way.

aimmo-game/simulation/action.py
PhydeauxLeChien: I take your point - name suggestions?

aimmo-game/simulation/action.py
```diff
@@ -1,145 +1,152 @@
...
-            respawn_location = world_map.get_random_spawn_location()
-            attacked_avatar.die(respawn_location)
-            world_map.get_cell(self.target_location).avatar = None
-            world_map.get_cell(respawn_location).avatar = attacked_avatar
-
-    def reject(self):
-        self.avatar.add_event(FailedAttackEvent(self.target_location))
-        self.avatar.clear_action()
+    def __init__(self, avatar, source, direction):
+        #                                               Untrusted data!
```
joshuablake: Typo?

aimmo-game/simulation/direction.py
```diff
@@ -16,6 +16,13 @@ def dict(self):
     def __repr__(self):
         return 'Direction(x={}, y={})'.format(self.x, self.y)
 
+    @staticmethod
+    def copy(other):
```
joshuablake: Directions are immutable so we shouldn't need to copy them.

aimmo-game/simulation/action.py
```diff
@@ -1,145 +1,152 @@
...
+        if not game_state.cell_occupied(self.target):
+            return self.apply(game_state, other_actions)
 
-        new_cell = world_map.get_cell(self.target_location)
-        if new_cell.pickup:
-            # TODO:  extract pickup logic into pickup when adding multiple types
-            self.avatar.health = min(10, self.avatar.health + new_cell.pickup.health_restored)
-            new_cell.pickup = None
+        next_avatar = game_state.avatar_at(self.target)
+        next_action = other_actions.by_avatar(next_avatar)
```
joshuablake: Fails if `other_actions is None` (I can't see any reason to assume not None here).

aimmo-game/simulation/action.py
```diff
@@ -1,145 +1,152 @@
...
-        self.avatar.add_event(FailedAttackEvent(self.target_location))
-        self.avatar.clear_action()
+    def __init__(self, avatar, source, direction):
+        #                                               Untrusted data!
+        Action.__init__(self, avatar, source, direction=Direction.copy(direction))
+
+    def _attacked_avatar(self, game_state, other_actions):
+        if game_state.cell_occupied(self.target):
+            return game_state.avatar_at(self.target)
+        elif (other_actions is not None
```
joshuablake: Can we pass in an empty ActionRegistry rather than having to check for None everywhere? This has a high probability of going wrong if someone later modifies the code and forgets [eg](https://github.com/ocadotechnology/aimmo/pull/153/files#diff-d04a139bc2877a24168490f27b3091a2R83).

aimmo-game/simulation/action.py
PhydeauxLeChien: Was trying to make clear which argument was untrusted data.

aimmo-game/simulation/direction.py
PhydeauxLeChien: they aren't actually, though they probably should be. Method name unclear in any case.

aimmo-game/simulation/action.py
```diff
@@ -1,145 +1,152 @@
...
+              and other_actions.num_moves_to(self.target) == 1):
+            return other_actions.moves_to(self.target)[0].avatar_id
```
joshuablake: `ActionRegistry.moves_to` returns a generator which doesn't support indexing (probably want a unit test for the case of attacking a moving avatar). Can use `next()` method to get the value.

aimmo-game/simulation/avatar/avatar_wrapper.py
```diff
@@ -38,7 +30,8 @@ def _construct_action(self, data):
         action_data = data['action']
         action_type = action_data['action_type']
         action_args = action_data.get('options', {})
-        action_args['avatar'] = self
+        action_args['avatar'] = self.user_id
+        action_args['origin'] = self.location
```
joshuablake: Non-obvious place to update due to [`origin` name change](https://github.com/ocadotechnology/aimmo/pull/153#discussion_r76992744).

aimmo-game/simulation/map_generator.py
```diff
@@ -17,11 +17,11 @@ def generate_map(height, width, obstacle_ratio=0.1):
     for x, y in shuffled(_get_edge_coordinates(height, width)):
         if (x, y) != (always_empty_edge_x, always_empty_edge_y) and random.random() < obstacle_ratio:
             cell = grid[x][y]
-            cell.habitable = False
+            cell._habitable = False
             # So long as all habitable neighbours can still reach each other,
             # then the map cannot get bisected
             if not _all_habitable_neighbours_can_reach_each_other(cell, world_map):
-                cell.habitable = True
+                cell._habitable = True
```
joshuablake: Why is habitable private if it's being modified (also [in `fog_of_war.partially_fog_cell`](https://github.com/ocadotechnology/aimmo/pull/153/files#diff-bd8d8b8b976f5cb3c4e2b1109294ed52R46)?

aimmo-game/simulation/direction.py
```diff
@@ -16,6 +24,13 @@ def dict(self):
     def __repr__(self):
         return 'Direction(x={}, y={})'.format(self.x, self.y)
 
+    @staticmethod
+    def from_dict(other):
+        try:
+            return Direction(**other.dict)
```
joshuablake: Looks like this is now unnecessary as `from_dict` is only called with actual dicts (which makes sense with the new name).

aimmo-game/simulation/action.py
PhydeauxLeChien: As discussed, you can assume it is not `None` here because of the conditions in `is_legal`

aimmo-game/simulation/location.py
```diff
@@ -1,10 +1,26 @@
...
 
     def __add__(self, direction):
-        return Location(self.x + direction.x, self.y + direction.y)
+        try:
+            d_x = direction.x
+            d_y = direction.y
+        except AttributeError:
+            if direction is None:
+                return self
+            raise TypeError('Not a direction.')
```
joshuablake: Should return `NotImplemented` in case the other object implements `__radd__`.

Also, should implement `__radd__` (so that `direction + location` works).

aimmo-game/simulation/turn_manager.py
```diff
@@ -79,42 +39,97 @@ def run_turn(self):
...
+        self._actions_by_target = {}
+
+    def by_avatar(self, avatar_id):
+        return self._actions_by_avatar.setdefault(avatar_id, None)
+
+    def avatar_moving(self, avatar_id):
+        return isinstance(self.by_avatar(avatar_id), MoveAction)
+
+    def _by_target(self, target):
+        return self._actions_by_target.setdefault(target, [])
```
joshuablake: IMO, cleaner to have `self._actions_by_target = collections.defaultdict(list)`.

aimmo-game/simulation/turn_manager.py
```diff
@@ -79,42 +39,97 @@ def run_turn(self):
...
+
+    def by_target(self, target):
+        return (action for action in self._by_target(target))
```
joshuablake: Why make a generator from a list rather than just returning the list?

aimmo-game/simulation/worker_manager.py
```diff
@@ -18,14 +18,12 @@ class _WorkerManagerData(object):
...
-
     def __init__(self, game_state, user_codes):
         self._game_state = game_state
         self._user_codes = user_codes
         self._lock = Semaphore()
 
     def _remove_avatar(self, user_id):
-        assert self._lock.locked
         self._game_state.remove_avatar(user_id)
         del self._user_codes[user_id]
```
joshuablake: Removing the lock means that a user can get added between these two lines leaving us in an inconsistent state: the user is in the game but we deleted their code.

aimmo-game/simulation/world_map.py
```diff
@@ -58,13 +57,28 @@ def __hash__(self):
...
 
+    @property
+    def avatar(self):
+        return self._avatar
+
+    @avatar.setter
+    def avatar(self, new_avatar):
+        if new_avatar is not None:
+            assert self.is_habitable, 'Cannot move avatar to inhabitable cell.'
+            assert self._avatar is None, 'Cannot move avatar to occupied cell.'
```
joshuablake: Throwing an `AssertionError` from an assignment statement is surprising for whoever is consuming this class's API. If we're doing checking I think it should be a `set_avatar` method. And probably throw a `ValueError`.

aimmo-game/simulation/world_map.py
```diff
@@ -87,66 +101,62 @@ class WorldMap(object):
...
-        if not self.is_on_map(location):
+        if not self.cell_on_map(location):
             raise ValueError('Location %s is not on the map' % location)
-        cell = self.grid[location.x][location.y]
+        cell = self._grid[location.x][location.y]
         assert cell.location == location, 'location lookup mismatch: arg={}, found={}'.format(location, cell.location)
         return cell
 
-    def clear_cell_actions(self, location):
+    def cell_on_map(self, location):
```
joshuablake: Should be `location_on_map`.

aimmo-game/simulation/world_map.py
```diff
@@ -87,66 +101,62 @@ class WorldMap(object):
...
-        return self.num_rows * self.num_cols
+            return False
 
-    def apply_score(self):
-        for cell in self.score_cells():
-            try:
-                cell.avatar.score += 1
-            except AttributeError:
-                pass
+    def cell_occupied(self, location):
```
joshuablake: This is never called by anything. I don't think we need to replicate the `Cell` interface here as you can always access that using `get_cell`.

aimmo-game/simulation/world_map.py
```diff
@@ -157,42 +167,49 @@ def _add_outer_layer(self):
...
                 cell.generates_score = False
 
-        new_num_score_locations = len(list(self.score_cells()))
+        new_num_score_locations = len(list(self.score_cells))
         target_num_score_locations = int(math.ceil(
             num_avatars * TARGET_NUM_SCORE_LOCATIONS_PER_AVATAR
         ))
         num_score_locations_to_add = target_num_score_locations - new_num_score_locations
-        locations = self._get_random_spawn_locations(num_score_locations_to_add)
+        locations = self._get_random_spawn_cells(num_score_locations_to_add)
```
joshuablake: `cells` not `locations`

aimmo-game/simulation/world_map.py
```diff
@@ -157,42 +167,49 @@ def _add_outer_layer(self):
...
             cell.generates_score = True
 
-    def _add_pickups(self, num_avatars):
+    def add_pickups(self, num_avatars):
         target_num_pickups = int(math.ceil(num_avatars * TARGET_NUM_PICKUPS_PER_AVATAR))
         LOGGER.debug('Aiming for %s new pickups', target_num_pickups)
-        max_num_pickups_to_add = target_num_pickups - len(list(self.pickup_cells()))
-        locations = self._get_random_spawn_locations(max_num_pickups_to_add)
+        max_num_pickups_to_add = target_num_pickups - len(list(self.pickup_cells))
+        locations = self._get_random_spawn_cells(max_num_pickups_to_add)
```
joshuablake: `cells` not `locations`

aimmo-game/simulation/world_map.py
```diff
@@ -157,42 +167,49 @@ def _add_outer_layer(self):
...
         for cell in locations:
             if random.random() < PICKUP_SPAWN_CHANCE:
                 LOGGER.info('Adding new pickup at %s', cell)
                 cell.pickup = HealthPickup()
 
-    def _get_random_spawn_locations(self, max_locations):
+    def potential_spawn_cells(self):
```
joshuablake: Should be private

aimmo-game/tests/test_service.py
```diff
@@ -35,9 +42,14 @@ def setup_world(self):
         ]
         grid = [[MockCell(Location(x, y), **CELLS[x][y])
                  for y in xrange(3)] for x in xrange(2)]
-        state_provider.set_world(GameState(WorldMap(grid), SimpleAvatarManager()))
+
+        __builtin__.turn_manager = TurnManager(
```
joshuablake: I think monkey patching would be better?
```python
import service
service.turn_manager = ...
```

aimmo-game/simulation/map_generator.py
PhydeauxLeChien: Both this and `partially_fog_cell` are modifying the attribute on what I would consider to be initialisation (either of a cell in the game state, or a copy in a view). Arguably these should be rewritten to decide whether the cell should be habitable before it is created, but the attribute should remain private imo.

aimmo-game/simulation/turn_manager.py
PhydeauxLeChien: Should return a "copy of the list" so that the internal structure of the registry won't be altered from outside - I use a generator in principle because of memory usage, though in this case it doesn't matter since the length of the list is at most 4.

aimmo-game/simulation/world_map.py
PhydeauxLeChien: Checks and validation in property setters are fine imo (it's basically what they're for). And I think asserts are fine for conditions that should *never* happen like these.

aimmo-game/tests/simulation/test_game_state.py
```diff
@@ -50,6 +51,7 @@ def test_add_avatar(self):
         self.assertIn(7, self.av_manager._avatars_by_id)
         self.assertEqual(self.av_manager._avatars_by_id[7].location, Location(10, 10))
 
+    @unittest.skip('Doesn\'t test fog of war.')
```
joshuablake: Think this should be `test_fog_of_war_applied` (ie: check that `GameState` calls `apply_fog_of_war` and then passes the result to `decide_action`).

aimmo-game/simulation/game_state.py
```diff
@@ -77,6 +78,16 @@ def active_avatars(self):
     def avatar_location(self, avatar_id):
         return self._avatar_manager.get_avatar(avatar_id).location
 
+    def avatar_at(self, location):
+        try:
+            return self._world_map.get_cell(location).avatar.user_id
+        except AttributeError:
+            return None
+
+    def decide_action(self, avatar_id):
```
joshuablake: I thought this was going on `DecisionThread` to remove the ability to change `GameState`?

PhydeauxLeChien: going to wait for another pull request for this, as it will require all the tests that use `DummyAvatar` to be rewritten completely.

aimmo-game/simulation/worker_manager.py
PhydeauxLeChien: Well the assert doesn't prevent that from happening either - reasoning for removing it was that it's a private method that only gets called inside a `with self._lock` block in other methods. Better than an assert would be to change `_lock` to a re-entrant lock and acquire it again.

Spycho:
Will cherry pick the bits we want, as it is rather stale and has lots of merge conflicts.

PR #205, opened 2017-07-13
fix #163 and other __ne__ issues

### LINKED ISSUE #163: test_map_contains_some_habitable_cell_on_border sporadically fails
We've got two problems:

- A non-deterministic test (which makes debugging hard and gives false results some of the time)
- An actual bug whereby we don't always guarantee a habitable cell on the edge (it usually has 1, but sometimes has 0).

For the first issue, I'd suspect either the use of un-seeded random, or iteration over a collection with non-deterministic iteration order.

Both need fixing, however if you make the test deterministic, you might not reproduce the second issue, and if you don't, it doesn't fail all of the time, and may not fail in the same way each time it fails, so it will be a pain to debug.

Spycho:
Location implements __eq__, but not __ne__, which means if comparing two locations that are the same `a == b` returns True, and `a != b` also returns True (because the default is object identity, rather than the negation of the __eq__ method. Sigh.

Fixed by implementing __ne__.

Also made map_generator tests deterministic (`random.seed(0)`).

### DIFF
--- a/aimmo-game-worker/simulation/location.py
+++ b/aimmo-game-worker/simulation/location.py
@@ -1,5 +1,4 @@
 class Location(object):
-
     def __init__(self, x, y):
         self.x = x
         self.y = y
@@ -15,6 +14,9 @@ class Location(object):
 
     def __eq__(self, other):
         return self.x == other.x and self.y == other.y
+
+    def __ne__(self, other):
+        return not self == other
 
     def __hash__(self):
         return hash((self.x, self.y))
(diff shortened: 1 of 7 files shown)


mikebryant:
Hmpf.
In Python3: `By default, __ne__() delegates to __eq__() and inverts the result unless it is NotImplemented. There are no other implied relationships among the comparison operators, for example, the truth of (x<y or x==y) does not imply x<=y. To automatically generate ordering operations from a single root operation, see functools.total_ordering().`

PR #136, opened 2016-08-10
Auth tokens

joshuablake:
Requires #124 as that adds the models in the backend [(diff with just this pull request)](https://github.com/joshuablake/aimmo/compare/multiple_games...joshuablake:auth_tokens).

Whenever a component receives a request, the receiver should check authentication from the caller. Workers should _never_ know any other components auth token due to running user code.

Places where needs/has been implemented:
- [x] players/creator interface (needs a new auth token generated before the creator app is started).
- [x] players/game interface (game auth token).
- [x] game/worker interface (avatar auth token).

PR #212, opened 2017-07-26
Front-end translated to Unity3D and multiple other changes

ramonfmir:
The main changes are:

- The 'watch' view is deployed to WebGL from Unity3D. This includes also health, scores and pickups.
- The API has been refactored to just sending updates (i.e. create/delete objects in the scene).
- Each avatar has its own view of the world and everything outside that view does not appear on the scene.
- Custom level generation from 2D maps.
- Enhanced current test suite.
- Added integration tests [not fully implemented].
- Support for multiple clients [not fully implemented].
- Better documentation throughout the code base.
- Other minor bug fixes.

Sorry about the massive pull request. In the future we'll do smaller ones.


Spycho:
Wow... This is a rather massive pull request. Ideally, it would be better if we could break pull requests up into smaller, useful but self contained changes, such that it is easier to review, and easier to make changes if some parts of it require it.

I haven't looked through the code yet, but something we'll need to address is the rather drastic drop in test coverage.

Spycho:
I think it's probably worth getting #150 and #136 merged in first, given the scale of the changes in this commit. @mikebryant might know more about those?

danalex97:
I looked at both coverage reports and it seems that not a lot of files have decreased. Also, the coverage is measured for the folder integration-tests/, which decreases it a bit. Another significant decrease is in service  and turn manager, which should be covered by the integration tests(when they are finished).

Is there a way to automatically compare coverage reports? I am looking at:
https://coveralls.io/builds/12560909
https://coveralls.io/builds/12540501

danalex97:
Regarding #150 and #136, we will probably look at them by hand as a lot of the logic is changed on unity.

danalex97:
The coverage reports seem to have been wrong. (#213) I will make a pull request for this issue. The difference in coverage seems to actually be 0.07%.

.gitignore
```diff
@@ -68,4 +68,4 @@ db.sqlite3
 example_project/example_project/static/
 
 # Created during minikube testing
-test-bin/
+test-bin/
```
mikebryant: Please don't remove newlines from ends of files

.coveragerc
```diff
@@ -0,0 +1,6 @@
+[report]
```
mikebryant: In all of the other python projects in Ocado, we use something like:
```
$ cat setup.cfg
[coverage:run]
source = django_autoconfig
```

Can we move this to the standard coverage config in setup.cfg?

.travis.yml
```diff
@@ -15,7 +15,7 @@ install:
 script:
   - python all_tests.py --coverage
 after_success:
-  - coveralls
+  - coveralls --rcfile=.coveragerc
```
mikebryant: Shouldn't need this

.travis.yml
```diff
@@ -24,5 +24,6 @@ deploy:
   distributions: "bdist_wheel sdist"
   on:
     repo: ocadotechnology/aimmo
+    all_branches: true
```
mikebryant: We do not want to deploy from all branches

aimmo-game-creator/service.py
```diff
@@ -7,9 +7,8 @@
 
 def main():
     logging.basicConfig(level=logging.DEBUG)
-    WorkerManagerClass = WORKER_MANAGERS[os.environ.get('WORKER_MANAGER', 'local')]
-    worker_manager = WorkerManagerClass(os.environ.get('GAME_API_URL',
-                                        'http://localhost:8000/players/api/games/'))
+    WorkerManagerClass = WORKER_MANAGERS[os.environ.get('WORKER_MANAGER', None)]
```
mikebryant: It doesn't really make sense to replace either of these with a default of `None` - it'll just break anyway.

If we truly don't want the default, we should just use `os.environ['WORKER_MANAHER']`, so it breaks with a nice error if we're missing the environment variable

aimmo-game-creator/worker_manager.py
```diff
@@ -171,7 +171,7 @@ def __init__(self, *args, **kwargs):
         super(KubernetesWorkerManager, self).__init__(*args, **kwargs)
 
     def _create_game_rc(self, id, environment_variables):
-        environment_variables['SOCKETIO_RESOURCE'] = "game/%s/socket.io" % id
```
mikebryant: Pretty sure changing this is wrong - it won't work under the reverse proxy on Kubernetes

aimmo-game/service.py
```diff
@@ -110,28 +123,39 @@ def player_data(player_id):
...
-
 def run_game(port):
     global worker_manager
 
     print("Running game...")
     settings = pickle.loads(os.environ['settings'])
-    api_url = os.environ.get('GAME_API_URL', 'http://localhost:8000/players/api/games/')
-    generator = getattr(map_generator, settings['GENERATOR'])(settings)
+
+    # TODO: this does not work with Kubernates; locally it works
```
mikebryant: This obviously needs to be fixed before merge

aimmo-game/service.py
```diff
@@ -110,28 +123,39 @@ def player_data(player_id):
...
 
-
 if __name__ == '__main__':
     logging.basicConfig(level=logging.INFO)
 
     socketio.init_app(app, resource=os.environ.get('SOCKETIO_RESOURCE', 'socket.io'))
+
     run_game(int(sys.argv[2]))
+
+    # run the flusk persistent connection
```
mikebryant: nit: flusk?

.travis.yml
danalex97: We did this in all AIMMO repos now. The coverage was taking tests and eggs into account.

aimmo-game/service.py
danalex97: This have been removed now. Only the comment remained.

aimmo-game-creator/worker_manager.py
danalex97: We only have one socket.io resource exposed. The problem came from the fact that we could not set the resource to game/%s/socket.io in Unity, so we changed the backend resource to be /socket.io

danalex97: Do you mean it won't work on Kubernates or Minikube?

aimmo-game-creator/service.py
danalex97: I agree, we will change that.

danalex97:
@mikebryant  Also, to let you know what we are changing now, when we talked to @Spycho last week we agreed on bringing Raphael JS so we can debug easier.  We also try to make sure localhost works with 2 players and minikube works with 2 players.

aimmo-game-creator/worker_manager.py
danalex97: Oh, I see.  I will read a bit about how this works. 

The reason why we wanted to change "game/%s/socket.io" to "/socket.io" is because we couldn't find out how to make it work like that from Unity.

ramonfmir:
@mikebryant I agree, we shouldn't have the Unity generated files in git. We can make ` run.py` clone [aimmo-unity](https://github.com/ocadotechnology/aimmo-unity/), build the project through a script and copy the files to make it work when we run it locally. Our idea was to do that once we incorporate the Unity repository to the current CI infrastructure, which will happen soon.

aimmo-game/service.py
danalex97: I removed the comment...

.travis.yml
danalex97: I think we still have problems with the actual coverage being not precise. Looking at the files in coveralls.io, I think this report is more accurate for master:

https://coveralls.io/github/ocadotechnology/aimmo?branch=coveralls

aimmo-game-creator/worker_manager.py
danalex97: @ramonfmir changed this back

PR #252, opened 2017-09-27
Minimum Product: Geography Package

OlafSzmidt:
`geography` package created within the game now. The following are moved to this:

- `direction`
- `cell`
- `location`

The moving of `direction` affected the back_end tests quite a lot. They all pass and the game seems to run fine (?), but do check just in case on your ends.


OlafSzmidt:
@CelineBoudier @mrniket @ramonfmir Any remarks?

Issue #245, opened 2017-09-26
Minimum Product: Porting world_map.py and test_world_map.py to py3.


OlafSzmidt:
I'm doing lots of refactoring of the codebase and getting rid of whatever we don't need for Level 1 minimum product. Stuff like fog of war or dynamic map expansion is going for now.

I'm constantly doing code analysis as I go along and this is the current result:

```
/home/olaf/IdeaProjects/forked/aimmo/aimmo-game/simulation/maps/world_map.py
    Warning:Warning:line (6)Unused import statement
    Warning:Warning:line (48)Python version 3.6 does not have method xrange
    Warning:Warning:line (49)Python version 3.6 does not have method xrange
    Warning:Warning:line (55)dict.iterkeys(), dict.iteritems() and dict.itervalues() methods are not available in py3
    Warning:Warning:line (181)Python version 3.6 does not have method xrange
    Warning:Warning:line (182)Python version 3.6 does not have method xrange
/home/olaf/IdeaProjects/forked/aimmo/aimmo-game/tests/test_simulation/test_world_map.py
    Warning:Warning:line (86)Python version 3.6 does not have method xrange
    Warning:Warning:line (86)Python version 3.6 does not have method xrange
    Warning:Warning:line (286)Python version 3.6 does not have method xrange
    Warning:Warning:line (286)Python version 3.6 does not have method xrange
    Warning:Warning:line (309)Python version 3.6 does not have method xrange
```

It's mostly cross-compatibility of python and we should fix this asap. If anyone else takes on this issue make sure you re-run the code-analysis before closing the issue.

OlafSzmidt:
#254 will solve this when merged.

PR #254, opened 2017-09-27
Minimum Product: World map ported to python3.

OlafSzmidt:
Ported `world_map` and `test_world_map` to python3, ie:

- `xrange` is now replaced is `range`.
- `try` and `except` used in instances where `itervalues()` is used. This way I can allow for both versions to run the code.

### DIFF
--- a/aimmo-game/simulation/world_map.py
+++ b/aimmo-game/simulation/world_map.py
@@ -2,7 +2,6 @@ import math
 import random
 from logging import getLogger
 
-from simulation.world_state import MapFeature
 from simulation.pickups import ALL_PICKUPS
 from simulation.location import Location
 from simulation.maps.cell import Cell
@@ -44,14 +43,23 @@ class WorldMap(object):
 
         (min_x, max_x, min_y, max_y) = WorldMap._min_max_from_dimensions(height, width)
         grid = {}
-        for x in xrange(min_x, max_x + 1):
-            for y in xrange(min_y, max_y + 1):
+        for x in range(min_x, max_x + 1):
+            for y in range(min_y, max_y + 1):
                 location = Location(x, y)
                 grid[location] = Cell(location)
         return cls(grid, new_settings)
 
     def all_cells(self):
-        return self.grid.itervalues()
+        '''
+        Written to support both python3 and python2. itervalues() became obsolete
+        in py3 and values() returns dictionary views, which are iterators.
+        :return: Iterator (dictionary or iterator) of all cells in the grid.
+        '''
+        try:
+            values = self.grid.itervalues()
+        except AttributeError:
+            values = self.grid.values()
+        return values
 
     def potential_spawn_locations(self):
         return (c for c in self.all_cells()
@@ -139,11 +147,11 @@ class WorldMap(object):
         self._add_horizontal_layer(self.max_y() + 1)
 
     def _add_vertical_layer(self, x):
-        for y in xrange(self.min_y(), self.max_y() + 1):
+        for y in range(self.min_y(), self.max_y() + 1):
             self.grid[Location(x, y)] = Cell(Location(x, y))
 
     def _add_horizontal_layer(self, y):
-        for x in xrange(self.min_x(), self.max_x() + 1):
+        for x in range(self.min_x(), self.max_x() + 1):
             self.grid[Location(x, y)] = Cell(Location(x, y))
 
     def _add_pickups(self, num_avatars):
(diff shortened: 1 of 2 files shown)


danalex97:
@OlafSzmidt From the stacktrace in #245, I guess you try to port the whole aimmo-game package. If so, you should also change the Dockerfile to python3.

OlafSzmidt:
@danalex97 There are hundreds of places where the 2to3 move needs to done, not just world_map. Small steps. Will the codebase work with py2 dockerfiles for now? I wanted this PR to just be for world_map functionality :)

danalex97:
@OlafSzmidt Oh, I see. That's why you did the itervalues to work with both versions. That makes sense.

OlafSzmidt:
@danalex97 Exactly ;) And itervalues() are not compatible with py3, but values() are back-compatible with py2 because they just return a list. I'm just pushing an extra test for that functionality if you could check that before you go :P

OlafSzmidt:
@danalex97 To be honest I was hoping that the `with self.assertRaises()` code block will force the `except` in the code to be ran. Coverage doesn't seem to show that it was covered though.... but......

```
self.assertFalse(cell_names, collections.Iterable)
self.assertTrue(cell_names, list)
```

Seems to pass; therefore I guess it's a problem with the coverage tool (or else I'm misunderstanding the use of `with self.assertRaises(Exception)`.

aimmo-game/tests/test_simulation/test_world_map.py
```diff
@@ -118,6 +120,22 @@ def test_all_cells(self):
         self.assertIn('D', cell_names)
         self.assertEqual(len(cell_names), 4)
 
+    def test_all_cells_python3(self):
+        # Except block will be called when python3 is used.
+        with self.assertRaises(Exception) as e:
```
danalex97: I think this test is wrong. This checks that an exception has been raised. The exception raised is in these lines:
```pyhton
self.assertFalse(isinstance(cell_names, collections.Iterable))
self.assertTrue(isinstance(cell_names, list))
``` 
since the assertion is false in python2. You can check this by commenting these lines. Thus, the test passes in python2. In python3, the test would fail since no exception would have been raised. 

I think we should have only one test, that should work regardless of the python version or no test at all.

aimmo-game/tests/test_simulation/test_world_map.py
```diff
@@ -305,7 +323,7 @@ class TestWorldMapWithOriginCentre(TestWorldMap):
     def _generate_grid(self, columns=2, rows=2):
         alphabet = iter(ascii_uppercase)
         grid = {Location(x, y): MockCell(Location(x, y), name=next(alphabet))
-                for x in xrange(-int_ceil(columns/2.0)+1, int_floor(columns/2.0)+1) for y in xrange(-int_ceil(rows/2.0)+1, int_floor(rows/2.0)+1)}
+                for x in range(-int_ceil(columns/2.0)+1, int_floor(columns/2.0)+1) for y in range(-int_ceil(rows/2.0)+1, int_floor(rows/2.0)+1)}
```
danalex97: Didn't notice. Good catch with those :+1:

aimmo-game/tests/test_simulation/test_world_map.py
OlafSzmidt: The previous test will pass on both versions of python; and I suppose test coverage isn't THAT important in this case. I'll get rid of it. Thanks! :+1:

danalex97: :+1:  I'm curious if that is something actually testable and how one would do that.

OlafSzmidt:
Thanks @danalex97. @CelineBoudier @ramonfmir @mrniket - any remarks before I merge?

aimmo-game/simulation/world_map.py
```diff
@@ -44,14 +43,23 @@ def generate_empty_map(cls, height, width, settings):
...
     def all_cells(self):
-        return self.grid.itervalues()
+        '''
+        Written to support both python3 and python2. itervalues() became obsolete
+        in py3 and values() returns dictionary views, which are iterators.
+        :return: Iterator (dictionary or iterator) of all cells in the grid.
+        '''
+        try:
+            values = self.grid.itervalues()
+        except:
```
CelineBoudier: you could try to except for a specific error type, less messy - in case the problem is not linked to the version of python for instance?

CelineBoudier:
but maybe squash before

OlafSzmidt:
Squashed.

PR #259, opened 2017-09-28
Minimum Product: State package

OlafSzmidt:
`game_state` and `world_state` is now in `simulation/state/.`.

PR #266, opened 2017-09-29
Minimum Product: State Package

OlafSzmidt:
Attempt 2 at cleaning the tree a little and pushing this state package. Should be alright to merge now. 

Fixes #259.


OlafSzmidt:
@CelineBoudier?

Issue #265, opened 2017-09-29
Port aimmo-game from py2to3. 


OlafSzmidt:
Port all simulation files, test files, levels files and service files to have python3 references. If possible, keep cross-compatibility.

OlafSzmidt:
[Something to consider.](http://python-future.org/automatic_conversion.html)

OlafSzmidt:
Implemented in #267 but we should definitely consider the thing in the link above instead at some stage.

PR #267, opened 2017-09-29
Ported game package to py3.

OlafSzmidt:
**Issue #265 fix:**

- Changed all instances of `xrange` to `range`.
- Found a use of `unicode` which is replaced by `str` (technically).

**Refactoring:**

- Removed an unused import.
- Changed a set function call in a test to a set literal. [Twice as fast execution.](https://renzo.lucioni.xyz/pythons-set-literals/)

### DIFF
--- a/aimmo-game/simulation/avatar/avatar_view.py
+++ b/aimmo-game/simulation/avatar/avatar_view.py
@@ -53,8 +53,8 @@ class AvatarView:
     @classmethod
     def cells_in_rectangle(self, top_left, bottom_right, world_map):
         cells = set([])
-        for x in xrange(max(top_left.x, world_map.min_x()), min(bottom_right.x, world_map.max_x() + 1)):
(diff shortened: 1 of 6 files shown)


CelineBoudier:
We need to try it with actual python 3 before :)

OlafSzmidt:
Don't think we can just yet. Packages such as `aimmo-worker` aren't ported yet. Python3 will work backwards but python2 specfic stuff won't work forward. `futurize` should be our priority instead of doing it manually perhaps.

danalex97:
Why does the fact that `aimmo-worker` is not ported yet affect `aimmo-game`?

OlafSzmidt:
@danalex97 - Hmm, not exactly sure. But `all_tests.py` will run tests from every directory, and if you run `run.py` on python3, it won't work just yet. 

Am I wrong? :~

danalex97:
@OlafSzmidt Probably not.

PR #276, opened 2017-10-03
Potential MockCell bug fixed. Fixed #274.

### LINKED ISSUE #274: Buggy constructor for MockCell in tests!
The `MockCell` constructor contains a parameter `actions` that has a default mutable argument. 

This has potential for bugs. If I understand correctly, a new list will not be created with each call that has no actions parameter specified. [Python's defaults arguments are evaluated once when the function is defined, not each time the function is called](http://docs.python-guide.org/en/latest/writing/gotchas/). 

Therefore later down the code when we actually mutate the actions list (which we believe is 'fresh'), we risk changing a list that was created for a previous `MockCell` instead. 

This only works correctly if we are only making one (or none) call to `MockCell` with `action` not specified.

OlafSzmidt:


OlafSzmidt:
@CelineBoudier @danalex97 ?

Issue #229, opened 2017-09-04
Add CodeClimate


danalex97:
Maybe we should add codeclimate as we are not consistent with the style everywhere.

CelineBoudier:
Yes I suggested it on @OlafSzmidt pull request :)

danalex97:
I know, I just thought we should add an issue to not forget about it.

CelineBoudier:
:+1:  Yes we should ^^ Thanks for adding this issue @danalex97

dionisioC:
Hi, this issuee is about add codeclimate, or refactor the codebase to be more pep8 compilant?

OlafSzmidt:
Hi @dionisioC - the issue is about adding codeclimate so that our code is more pep8 compliant.

OlafSzmidt:
Changed the title and description so we don't confuse anyone else ;)

dionisioC:
Thanks, I forked the repo, I want to try to add codeclimate

OlafSzmidt:
@dionisioC Good luck!

CelineBoudier:
Relevant

CelineBoudier:
Added the dashboard

PR #293, opened 2017-10-05
Add code climate #229

dionisioC:
This is the config needed (mainly) to configure code climate.

GPA and issues are working as you can see here

https://github.com/dionisioC/aimmo/tree/Add_CodeClimate_%23229

Coverage needs to be configured in travis (done) but I can't test it

The url for the badges must be changed for the ones of the official account


dionisioC:
The build is mistakenly passing, seems like the result of the coverage analysis are not found, do you know where are them?

danalex97:
@dionisioC Maybe you need to add some other flags in the `all_test.py` file. For example [here](https://pypi.python.org/pypi/python-coveralls/) they have a `--pep8` flag. Sorry if is not this, I don't have much experience with this kind of stuff.

dionisioC:
In theory that is handled by a codeclimate engine (.codeclimate.yml)
  pep8:
    enabled: true

A file with the test coverage analisys should be generated in some part, but I am not able to find it

codeclimate GPA and issues are working, coverage is not yet.

I will check that. Thanks for the help!

OlafSzmidt:
Any updates @dionisioC?

dionisioC:
No, I haven't have time to check @danalex97 suggestion

dionisioC:
Last 3 thing I tested didn't work

dionisioC:
I added debug , bu t I hve no clue on what is waiting for, if there is no clue on how to fix this I will close the pull request so other person can do it

PR #305, opened 2017-10-19
Revert to a fixed state

OlafSzmidt:
As discussed with @Spycho, @CelineBoudier and everyone else; we are reverting the game to the last known working state.

Could everyone please, test this if the game works properly? Use the test plan to see if the game works properly. It would be nice to see this being tested on both Windows and Linux.

@CelineBoudier, can you please add a tag to the head of the current master so it will be easily recognizable? Github doesn't support tag merging via a PR.


OlafSzmidt:
Currently a simple test plan can be found [here](https://docs.google.com/a/ocado.com/document/d/1zycqVEp9_yKRJivKyumk1S7fjQwXIrSlaHUVdTf5sCw).

Issue #369, opened 2017-11-17
Better color functionality 


OlafSzmidt:
A comment made in `player_dict` of `service.py` in `aimmo-game` states:

`TODO: implement better colour functionality: will eventually fall off end of numbers`

The value in question is the `bodyFill` of `colours` which right now does:

`"#%06x" % (avatar.player_id * 4999)`

I'm removing this in the next PR and making note of it here. 

**Perhaps we can remove this functionality altogether**?

OlafSzmidt:
Not applicable right now.

PR #370, opened 2017-11-17
Minor refactoring

### LINKED ISSUE #368: Score_cells use tuples. That might be too slow!
`Score_cells()` function in `world_map.py` uses a tuple. If I recall correctly this will be very slow and can be improved.

New starter issue perhaps?

OlafSzmidt:
Sticking to our **new 90 character rule** and slowly changing the code to suit this.

Also fixed #368, but not because of the initial reason we thought of which was speed. In fact tuples have a slight edge of lists in this scenario, but the convention states that for homogeneous collections lists should be used instead.


aimmo-game/service.py
```diff
@@ -107,9 +106,13 @@ def run_game(port):
     generator = getattr(map_generator, settings['GENERATOR'])(settings)
     player_manager = AvatarManager()
     game_state = generator.get_game_state(player_manager)
-    turn_manager = ConcurrentTurnManager(game_state=game_state, end_turn_callback=send_world_update, completion_url=api_url+'complete/')
+    turn_manager = ConcurrentTurnManager(game_state=game_state,
```
mrniket: so nice :)

aimmo-game/simulation/map_generator.py
```diff
@@ -48,15 +48,16 @@ def get_map(self):
...
         world_map = WorldMap.generate_empty_map(height, width, self.settings)
 
-        # We designate one non-corner edge cell as empty, to ensure that the map can be expanded
+        # We set one non-corner edge cell as empty, to ensure that the map can be expanded
         always_empty_edge_x, always_empty_edge_y = get_random_edge_index(world_map)
         always_empty_location = Location(always_empty_edge_x, always_empty_edge_y)
 
         for cell in shuffled(world_map.all_cells()):
-            if cell.location != always_empty_location and random.random() < self.settings['OBSTACLE_RATIO']:
+            if cell.location != always_empty_location and\
```
mrniket: In my code I have always put the `and` on the new line, is there a python convention for where it should go?

aimmo-game/simulation/world_map.py
```diff
@@ -221,7 +234,8 @@ def _reset_score_locations(self, num_avatars):
             cell.generates_score = True
 
     def _add_pickups(self, num_avatars):
-        target_num_pickups = int(math.ceil(num_avatars * self.settings['TARGET_NUM_PICKUPS_PER_AVATAR']))
+        target_num_pickups = int(math.ceil(num_avatars *
```
mrniket: if you make a variable for the `num_avatars * self.settings['TARGET_NUM_PICKUPS_PER_AVATAR']` it might be a better way of splitting it up

aimmo-game/tests/functional/test_movements_in_map.py
```diff
@@ -14,15 +16,19 @@ class TestMovementsInMap(TestCase):
         'OBSTACLE_RATIO': 0,
     }
 
-    def set_up_environment(self, dummy_list=None, location=Location(0, 0), map_generator_class=map_generator.Main):
+    def set_up_environment(self, dummy_list=None, location=Location(0, 0),
```
mrniket: maybe we can make it consistent and have a new line per parameter?

aimmo-game/simulation/map_generator.py
OlafSzmidt: `and` stays on the first line. It was changed in 2016, I seen it in the changelogs of PEP8 today. Also, their official example guideline class, `Rectangle` shows this:

http://pymbook.readthedocs.io/en/latest/pep8.html

CelineBoudier: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

OlafSzmidt: Found it! https://hg.python.org/peps/log/tip/pep-0008.txt

More specifically: https://hg.python.org/peps/rev/502239ce4889

aimmo-game-worker/simulation/world_map.py
```diff
@@ -46,13 +46,13 @@ def all_cells(self):
         return self.cells.values()
 
     def score_cells(self):
-        return (c for c in self.all_cells() if c.generates_score)
+        return [c for c in self.all_cells() if c.generates_score]
```
CelineBoudier: might want it to stay a tuple - don't need to modify the results anywhere

aimmo-game/simulation/map_generator.py
OlafSzmidt: But we should avoid **line breaks** unless required, and instead use parenthesis for stuff like this to split it over few lines... I'll change it!

aimmo-game/tests/functional/test_movements_in_map.py
OlafSzmidt: I considered this but it seems like a bit of overkill, I only go to the next line when I run out of 90 character limit. 

cc @CelineBoudier watchu say

aimmo-game/service.py
```diff
@@ -46,7 +44,7 @@ def player_dict(avatar):
         'rotation': 0,
         "colours": {
             "bodyStroke": "#0ff",
-            "bodyFill": colour,
+            "bodyFill": "#%06x" % (avatar.player_id * 4999),
```
CelineBoudier: a bit dangerous or is it just me?

aimmo-game/simulation/world_map.py
OlafSzmidt: I kind of lost consistency here, I'll commit it like the rest of the file does it and try to keep this standard. Comment if you still don't like it ;p

aimmo-game/simulation/map_generator.py
CelineBoudier: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

aimmo-game/tests/functional/test_movements_in_map.py
CelineBoudier: i see your point @mrniket but it might indeed be an overkill. If there are a lot of parameters, it doesn't improve the readability IMO

aimmo-game/service.py
OlafSzmidt: How is it different to the one we had before? I did open an issue about this though #369

CelineBoudier: it's not different - was dangerous too

OlafSzmidt:
Also: https://github.com/ocadotechnology/aimmo/wiki/Conventions

aimmo-game/tests/functional/test_movements_in_map.py
mrniket: fair enough, for me it improves readability if there are more than a certain number of parameters but if it's just me then I guess it's okay 👍 . Also if there are a lot of parameters I would argue splitting the function up or grouping the parameters together

aimmo-game/service.py
mrniket: I wouldn't worry too much about this if we are going to start using the 3d models, the colour is not passed to the client anyways. Actually that means that this can probably be removed

OlafSzmidt: Should I make an issue @mrniket?

mrniket: let's add something onto #369

PR #386, opened 2017-11-24
First integration test written

OlafSzmidt:
Things to note:

~~I created an issue on github as this will not run with `all_tests.py` just yet. It's a process specific problem I can't solve right now. When running the file separately everything finishes and terminates, when the test runs via `all_tests.py`, the server keeps running and no requests are sent.~~

** ~~Python might not be killed correctly.~~**

# Fixed!


CelineBoudier:
I would maybe have made another pull request just for the refactoring to merge that first

integration-tests/tests/connection_set_up.py
```diff
@@ -0,0 +1,102 @@
...
+    :return: Response of POST.
+    """
+    response = None
+
+    logging.debug("Attempting a POST request for the provided URL...")
+
+    try:
+        response = session.post(url, data=data)
+        response.raise_for_status()
+    except requests.exceptions.RequestException as e:
```
mrniket: are we sure that we want to catch this? by the time this is called the server should be up, if it goes down we want the report this?

mrniket: perhaps we can use the `logging` package here

setup.py
```diff
@@ -24,6 +24,7 @@
     tests_require=[
         'django-setuptest',
         'httmock',
+        'lxml',
```
mrniket: where is this used?

OlafSzmidt:
@CelineBoudier Do you not think it's a bit of an overkill to create a new PR for 3 new line breaks? I see where you're coming from though, as refactoring code can get things messy quickly - just not for things like this imo

setup.py
OlafSzmidt: Good catch, I forgot to remove it

integration-tests/tests/connection_set_up.py
OlafSzmidt: This isn't a test so yes, I think we should catch it. Tests will have a HTTP code assertion to see if everything was successful after this POST request is made. 

Instead of failing logic by getting random exceptions, I think we should fail it by getting a different response than expected imo.

Issue #430, opened 2018-01-18
Refactor tearDown() in test_kubernetes


OlafSzmidt:
Seems like `tearDown()` in these tests will be repeating code quite a lot, they all do the same thing: receive the array of processes and kill the children, followed by the parent processes. Refactoring.

OlafSzmidt:
Done already in #431.

Issue #429, opened 2018-01-18
Add an option/flag to not print (ie. pipe) the text output of runner


OlafSzmidt:
All kubernetes test will be using the runner function introduced today. Every time these tests run, our travis logs will get spammed with all the output. I think we should disable this by a flag which can be changed for the purposes of debugging locally.

OlafSzmidt:
Implemented in #431

PR #431, opened 2018-01-18
Kubernetes Tests & Architectural Changes

OlafSzmidt:
### Changes made:

- `docker-py` updated to `docker`, API changed slightly.
- `travis.yml` now takes care of downloading our `kubectl` and `minikube` dependancies, not the script in the python backend.
- Packaged game runner files into `aimmo_runner` (note underscore)
- Refactoring into `shell_api` for shell methods.
- **Runner from now on should be ran as a function, not as a file.** It returns an array of processes that need to be killed. It also has flags which can be passed in.
- Skeleton for kubernetes integration tests is now there.
- Bits and bobs needed to make it work
- **UPDATE:** Capturing output is now possible for running the servers in the test, both async and sync.

### TODO:
- ~Pipe output so it's not printed from current integration tests. Perhaps this should be an extra flag in runner? #429 is open.~
- Seems like `tearDown()` in these tests will be repeating code quite a lot, they all do the same thing: receive the array of processes and kill the children, followed by the parent processes. Refactoring. #430
- Test interfaces are left for what we think should be the next tests that need to be written for this code, and one is skipped by a decorator `unittest.skip` because we can't get it to work and it becomes very time consuming ;( Issue #432


aimmo_runner/minikube.py
```diff
@@ -2,82 +2,13 @@
...
-def binary_exists(filename):
-    # Check if binary is callable on our path
-    try:
-        run_command([filename], True)
-        return True
-    except OSError:
-        return False
-
-
-def download_kubectl():
```
mrniket: we should change the readme to signify that you have to download and install kubectl and minikube and a vm thing yourself

aimmo_runner/runner.py
```diff
@@ -0,0 +1,75 @@
...
+    run_command(['python', _MANAGE_PY, 'migrate', '--noinput'])
+    run_command(['python', _MANAGE_PY, 'collectstatic', '--noinput'])
+
+    create_superuser_if_missing(username='admin', password='admin')
+
+    server_args = []
+    if use_minikube:
+        # Import minikube here, so we can install the deps first
+        parent_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+        sys.path.append(os.path.join(parent_dir, "aimmo_runner"))
```
mrniket: single quotes here?

integration-tests/tests/test_kubernetes.py
```diff
@@ -0,0 +1,157 @@
...
+    def test_adding_custom_game_sets_cluster_correctly(self):
+        """
+        Log into the server as admin (superuser) and create a game
+        with the name "testGame", using the default settings provided.
+        """
+
+        def _wait_for_kubernetes_cluster(api_instance):
+            time_elapsed = 0
+
+            while time_elapsed <= 200:
```
mrniket: I would rename this to timeout

minikube_requirements.txt
```diff
@@ -1,2 +1,2 @@
-docker-py >= 1.10
+docker >= 1.10
```
mrniket: I think we should bump the min version on this

setup.py
```diff
@@ -20,7 +20,8 @@
         'six',
         'pykube',
         'hypothesis',
-        'flask-cors'
+        'flask-cors',
+        'psutil',
```
mrniket: can we fix the major and minor version on this?

aimmo_runner/minikube.py
OlafSzmidt: Done 👍

aimmo_runner/setup.cfg
```diff
@@ -0,0 +1,13 @@
+[bdist_wheel]
+universal = 1
+
+[pep8]
+max-line-length = 160
+
+[versioneer]
+VCS = git
+style = pep440-pre
+versionfile_source = ../players/_version.py
```
CelineBoudier: hmm could that be set up elsewhere so that it's less susceptible to relative path changes?

README.md
```diff
@@ -80,8 +80,8 @@ Even with these basic mechanics, there is quite a lot of complexity in creating
     * All platforms: VT-x/AMD-v virtualization.
     * Linux: [Virtualbox](https://www.virtualbox.org/wiki/Downloads).
     * OSX: either [Virtualbox](https://www.virtualbox.org/wiki/Downloads) or [VMWare Fusion](http://www.vmware.com/products/fusion.html).
+* Download Docker, Minikube, and Kubectl before running the script.
```
CelineBoudier: can we make it easier for people that are working locally though?

aimmo_runner/setup.cfg
OlafSzmidt: We have a root directory dependant on where the project is ran. If it's local, then it's your typical path, but Travis is different, so this is checked at the start and saved in the `runner.run` file. We could use that variable here, but it would take a lot of work to get it passed from python to a `.cfg` file. If you think it's important I could start looking into it?

aimmo_runner/runner.py
OlafSzmidt: 👍

README.md
mrniket: I think this is the command you can run for mac providing you have homebrew installed: 
`brew update && brew install kubectl && brew cask install docker minikube virtualbox`

OlafSzmidt: Done for Mac. Can't do it for other platforms as I can't test

OlafSzmidt:
- **UPDATE:** Capturing output is now possible for running the servers in the test, both async and sync.

aimmo_runner/setup.cfg
OlafSzmidt: More things to maintain though ;( I'll see if we can maybe test this, or if it's already done

README.md
OlafSzmidt: Done 👍

CelineBoudier:
the name of the branch :joy: cc @danalex97

OlafSzmidt:
loool

aimmo_runner/setup.cfg
mrniket: Could you write an issue for this?

OlafSzmidt: #435

PR #444, opened 2018-01-29
Propagate settings down from django to unity for staging

OlafSzmidt:
- `http` or `https` not included in the propagaed URL down to unity. 
- Added logic to get the game port.
- We now pass a SSL flag instead of setting it in World Controller.

### DIFF
--- a/players/templates/players/watch.html
+++ b/players/templates/players/watch.html
@@ -15,17 +15,17 @@
 <script type="text/javascript">
 var GAME_URL_BASE = "{{ game_url_base }}";
 var GAME_URL_PATH = "{{ game_url_path }}";
+var GAME_URL_PORT = {{ game_url_port }};
+var GAME_SSL_FLAG = "{{ game_ssl_flag }}";
+
 
 // Send host and port to Unity.
-var path = GAME_URL_PATH;
-var proc = (GAME_URL_BASE.substr("http://".length)).split(":");
-var host = proc[0];
-
 function SendAllConnect() {
-  console.log(`Sending Messages to ${host}${path}`);
-  gameInstance.SendMessage("World Controller", "SetGameURL", host);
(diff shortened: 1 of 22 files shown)


players/templates/players/watch.html
```diff
@@ -15,17 +15,17 @@
...
-var proc = (GAME_URL_BASE.substr("http://".length)).split(":");
-var host = proc[0];
 
+// Send host and port to Unity.
 function SendAllConnect() {
-  console.log(`Sending Messages to ${host}${path}`);
-  gameInstance.SendMessage("World Controller", "SetGameURL", host);
-  gameInstance.SendMessage("World Controller", "SetGamePort", 0);
-  gameInstance.SendMessage("World Controller", "SetGamePath", path);
+  console.log(`Sending Messages to ${GAME_URL_BASE}${GAME_URL_PATH}`);
```
mrniket: this log message could be confusing as it contains partial information (only the url base and the path) we need to have a better log message here

players/tests.py
```diff
@@ -6,7 +6,9 @@
 
 from players import models, views
 
-views.app_settings.GAME_SERVER_LOCATION_FUNCTION = lambda num: ('base %s' % num, 'path %s' % num)
+views.app_settings.GAME_SERVER_URL_FUNCTION = lambda num: ('base %s' % num, 'path %s' % num)
+views.app_settings.GAME_SERVER_PORT_FUNCTION = lambda num: 0
```
mrniket: can we rename num to game_id?

players/tests.py
```diff
@@ -6,7 +6,9 @@
...
+views.app_settings.GAME_SERVER_SSL_FLAG = lambda ssl: True
```
mrniket: I think this just needs to be a boolean variable, not a function

PR #450, opened 2018-01-31
Create a deploy script that runs locally

OlafSzmidt:
Deploy script that alters GCP clusters! You need to authenticate locally to run it. Download `google-cloud` to do so.

### DIFF
--- a/deploy_dev.py
+++ b/deploy_dev.py
@@ -1,0 +1,76 @@
+from google.cloud import container_v1
+import os
+import kubernetes
+import yaml
+
+# THIS WILL BREAK WHEN THE FILE IS MOVED. IT LOOKS FOR ROOT OF PROJECT.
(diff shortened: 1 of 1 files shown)


deploy_dev.py
```diff
@@ -0,0 +1,76 @@
...
+
+# Load the default service account. Should be authenticated locally by the above
+# login command.
+kubernetes.config.load_kube_config()
+api_instance = kubernetes.client.CoreV1Api()
+extensions_api_instance = kubernetes.client.ExtensionsV1beta1Api()
+
+# TODO later: "kubectl cluster-info" might be useful here to check status?
+
+# TODO later: restart ingress
```
mrniket: we do this already

mrniket: so we can remove the comment :)

PR #455, opened 2018-02-01
Temporary deployment solution & documentation update

OlafSzmidt:
~~This PR will add a temporary deployment script and a relevant .yaml file which we can run manually until we manage to debug and set the appengine one correctly.~~

~~Reason for it being that if we make changes on appengine one and they don't work, we can still deploy easily using this.~~

~~To be deleted when appengine deploys automatically.~~

## Documentation has to be reviewed here, as it's in the wiki:
screen shot 2018-02-01 at 15 27 23

screen shot 2018-02-01 at 15 28 29

### DIFF
--- a/deploy_dev.py
+++ b/deploy_dev.py
@@ -2,24 +2,29 @@ from google.cloud import container_v1
 import os
 import kubernetes
 import yaml
+import requests
 
-# THIS WILL BREAK WHEN THE FILE IS MOVED. IT LOOKS FOR ROOT OF PROJECT.
+# THIS IS A TEMPORARY LOCAL SCRIPT TO DEPLOY INGRESS ON DEV. WHEN AUTOMATIC
+# DEPLOYMENT ON APP-ENGINE IS ADDED. THIS CAN BE DELETED.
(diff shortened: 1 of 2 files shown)


deploy_dev.py
```diff
@@ -2,25 +2,30 @@
 import os
 import kubernetes
 import yaml
+import requests
 
-# THIS WILL BREAK WHEN THE FILE IS MOVED. IT LOOKS FOR ROOT OF PROJECT.
+# THIS IS A TEMPORARY LOCAL SCRIPT TO DEPLOY INGRESS ON DEV. WHEN AUTOMATIC
+# DEPLOYMENT ON APP-ENGINE IS ADDED. THIS CAN BE DELETED.
```
CelineBoudier: second full stop could be a comma or plain deleted?

OlafSzmidt:
Added screenshots for docs.

OlafSzmidt:
Do not merge ❌

OlafSzmidt:
GCE deployment seems to be working, so I guess we no longer need to merge this. Unless we want to have a manual one at end to maintain too?

OlafSzmidt:
Closing as we no longer need a manual script.

Issue #586, opened 2018-05-17
Sort out docker tags & versioning


OlafSzmidt:
Quite a big task but will save hassle in the future. Right now we only deploy one tag to dockerhub. For instance, look at `aimmo-game-creator`, here:

https://hub.docker.com/r/ocadotechnology/aimmo-game-creator/

The hub only hosts one tag, `latest`. This is fine, and we deploy there every time we merge something into master with the tag `latest`. Locally we build `test` tags, and use these docker images to run a local testing environment.

The problem is with our dev and staging environments. They always pull from `latest` with the same code that production uses. This means that if we want to test any yaml changes, then we have to merge into master each time and redeploy with the new docker image.

PR #587, opened 2018-05-17
Further optimisations of k8s

OlafSzmidt:
This will be an ongoing process because of #586 unfortunately. Until that is solved there will be many PR's coming in.

From the statistics I've seen, these should work fine for now.

Another problem we have is the amount of nodes we run in parallel. They're not very powerful CPU wise, and standalone k8s already takes about half of the mCPU available. We run things like heapster on there. Solution would be to get more powerful individual machines, not scale sideways. Plus loads of other things, but that's for later ;)

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/587)

### DIFF
--- a/aimmo-game/simulation/worker_manager.py
+++ b/aimmo-game/simulation/worker_manager.py
@@ -256,6 +256,10 @@ class KubernetesWorkerManager(WorkerManager):
                                 'cpu': '10m',
                                 'memory': '64Mi',
                             },
+                            'requests': {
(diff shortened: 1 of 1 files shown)


mrniket:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/587)*

PR #657, opened 2018-07-11
Error catching & worker creation changes

OlafSzmidt:
These are mainly changes to how our workers (in local; threads, in k8s mode; pods) react to code changes. We had some discussion to see if we can remove the recreation of these pods completely but the final decision is that we do not want to persist memory and recreation actually makes sense.

The recent code had a bug where after the flask microservice restarted for the `aimmo-game-worker` after new code was POSTed; the new spawn location would be random. 

When we change code now, the avatar still exists in the game, but we restart the pod that it's running in. This way turns can still be processed and we will always have a location one away. 

**Changes in PR:**
- `aimmo-game-worker`'s `service.py` file now doesn't accept `directory`. Options are not used anywhere and we are slowly moving towards removing legacy code.
- `service.py` of `aimmo-game-worker` now imports the code in `process_turn()`, not when the worker is initialised in `run`. This way we can wrap it in try except blocks and return a `WaitAction()` when there is an issue caught. 
- Tests by @riaJha97 and @mrniket. **Still waiting for more before merging.**
- Docs by @NiallEgan 
- Status error catching in `_fetch_action` before deserialising.
- Major refactoring of worker management in `worker_manager.py`

**Issues:**
- ~There is a buffering issue. Each time the flask server restarts, we set the Unity game a few turns behind. With enough code POSTing, eventually the gap will be too big. @JCobbles started some work on syncing this when POSTing new code.~ https://github.com/ocadotechnology/aimmo-unity/pull/222

**Possible future work:**
- Do some static python code compiling before it hits the try except block for some optimisation? Perhaps we don't need to restart the flask server if we know the code will be incorrect?

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/657)


aimmo-game/simulation/worker_manager.py
```diff
@@ -33,16 +34,15 @@ def _remove_avatar(self, user_id):
         self._game_state.remove_avatar(user_id)
         del self._user_codes[user_id]
 
+    def is_new_avatar(self, user):
+        with self._lock:
+            existing_code = self._user_codes.get(user['id'], None)
+            return existing_code is None
+
     def remove_user_if_code_is_different(self, user):
```
NiallEgan: Maybe rename this method since it doesn't actually remove the user anymore?

aimmo-game/simulation/communicator.py
```diff
@@ -0,0 +1,15 @@
+import requests
+
+class Communicator():
```
NiallEgan: I feel like the purpose of this class isn't 100% clear, what is it communicating with? Maybe improve its name or add a docstring

PR #699, opened 2018-07-31
Back-end python socket tests & minor refactoring

OlafSzmidt:
## This is a PR into `feature/guessLogs`.

- Add tests.
- Now ignoring vscode local files.
- Renames `socketio` from `service` to `socketioserver`. `mock.patch()` had an issue determining between the two because of an alias, and it was resolving to the package rather than the local scope variable.
- Minor refactoring of how we regex find the avatar_id from query.
- Some error recording for incorrect query_data.

Things to consider:

- ~If you look at `test_send_world_update_for_one_user` and `test_send_world_update_for_multiple_users` they expect "Testing" as the contents of the `logs` key now. This was surprising because I expected the get call to `logs_provider` to fail in service and `''` to be returned as `avatar_logs`. Turns out that's not the case. The only place the "Testing" string gets inserted as logs is inside the `DummyAvatar` which is not used in this test. I can't debug this and figure out why this is happening. Test is good as it is but it's worrying leaving it unexplained.~ **This is explained in https://github.com/ocadotechnology/aimmo/issues/705 now.**

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/699)


OlafSzmidt:
Added a TODO that I'm still planning to change here. Will change these minor things soon.

OlafSzmidt:
Blocked by #705.

OlafSzmidt:
No longer blocked.

PR #784, opened 2018-09-04
Upgrade babel 7

mrniket:
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/784)


mrniket:
So parcel doesn't support babel yet. I'm going to split this up into linting fixes and a babel upgrade

Issue #804, opened 2018-09-10
Printing non-ascii characters crashes aimmo


dent50cent:
<b>Describe the bug</b>
When using print within aimmo, it is possible to attempt to print non-ascii characters such as emoji. However when this code is run the game crashes, opening the console shows a 401 error. The game remains broken even after the character has been removed and the code rerun.

<b>To reproduce</b>
1. Go to the aimmo game page
2. Create a print statement that uses a non-ascii characters
3. Observe that the game stops, and upon opening the console you should see a 401 response by the server.

<b>Expected behaviour</b>
The game should not crash when trying to print non-ascii characters. It should either print them (requires python 3+), or inform the user that they cannot print non-ascii characters.

<b>screenshots</b>
![Screen Shot 2018-09-10 at 11.44.52.png](https://images.zenhubusercontent.com/5b8f9b976cb9d0508bb5f54d/871b8f9f-fb98-47b2-b4b9-f597b2afe222)
![Screen Shot 2018-09-10 at 11.54.10.png](https://images.zenhubusercontent.com/5b8f9b976cb9d0508bb5f54d/88c27b4d-0d56-49d9-b89b-b454c768a9da)

OlafSzmidt:
This is not a bug. Django already takes care of our encoding. You should be able to print emoji's fine, even on python2. I'm running `Python 2.7.15rc1` and in local mode it's printing fine as long as you remove a redundant log line that was left in the code.

We encode our strings using UTF-8. As far as I know you need to explicitly convert ("decode" in python's terms, confusing) the string to unicode using UTF-8 otherwise you get what happened here.

![screenshot from 2018-09-12 23-00-14](https://user-images.githubusercontent.com/22431381/45456021-15890c00-b6e1-11e8-8229-824ba8e19092.png)
![screenshot from 2018-09-12 23-01-39](https://user-images.githubusercontent.com/22431381/45456022-15890c00-b6e1-11e8-88b7-1584fa6d42c9.png)

Will make a quick PR for this and then we can let users use more than just ASCII (which is what should happen imo).

PR #818, opened 2018-09-12
Fix 804, encoding user code

OlafSzmidt:
Fixes the issue described in #804. We can let users use non ASCII (ie. UTF-8 encoded unicode) text fine imo. It was just a cast to str that wasn't being encoded in the same format as the rest.

I also refactored the boolean logic. `None` evaluates to false and so does an empty string :)

Everything I said might be wrong in k8s mode though. I don't know what to expect in terms of string encoding on each of the pods.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/818)


OlafSzmidt:
There seems to be an issue with the snyk manifests?

OlafSzmidt:
I don't want to merge this as I don't have access to deployment servers etc. If someone could please do it ;)

PR #832, opened 2018-09-25
created refactor doc

dent50cent:
this doc outlines the solutions we've come up with to solve the circular dependency issues within our projects.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/832)


dent50cent:
Document does not need to be merged in, and is no longer needed

PR #962, opened 2019-01-29
replace handle_turn with next_turn

dent50cent:
Default code now looks like this:
```python
class Avatar:
    def next_turn(self, world_state, avatar_state):
        return MoveAction(direction.NORTH)
```
as per issue requirements. I kept the functionality for the original handle_turn method as well so existing games will still function for now as it was a relatively simple tweak to the code.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/962)


j4mesholland:
LGTM

Issue #1011, opened 2019-04-03
Delete and recreate game pods based on game status


faucomte97:
## Task Description
The game pods should be deleted after the game has been inactive for one hour (ie status is STOPPED).
The game pods should then be recreated whenever the game becomes active again (status is RUNNING) - which would occur as soon as any of the game's players loads the game.

## Acceptance Criteria
- [ ] Game pods effectively deleted after a game is STOPPED.
- [ ] Game pods effectively recreated after a game is RUNNING again.

## Analytics Requirements
N/A

PR #1026, opened 2019-04-18
Authenticate games

dent50cent:
## Description
*Required by #1011*

There is now a way to make sure a request is coming from a game using a simple token system.

At the first opportunity during the game creation, a token is generated and sent to the Django server. Then in future requests the game can use this token to prove who it is.

#### Things to be aware of

* There is a small amount of time between when the object is added to the database and when the token is recieved. (dependant on pod startup time)
* When redeploying games do not keep their old tokens.
* token is a urlsafe string of random charaters and symbols

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1026)


mariaFernando:
___
*[aimmo/permissions.py, line 28 at r1](https://reviewable.io/reviews/ocadotechnology/aimmo/1026#-Lcjd_RV-ZXnPAXUSVEs:-Lcjd_RW1ZZPjdnNhSA-:b-fkm8wr) ([raw file](https://github.com/ocadotechnology/aimmo/blob/e02e3e3afc45c28d05b0d8a15b717e96037919e5/aimmo/permissions.py#L28)):*
> ```Python
>         if not self.token_requested:
>             self.token_requested = True
>             print(self.token_requested)
> ```

Is this print statement necessary?

PR #1035, opened 2019-05-16
Testing pod stop issues

dent50cent:
## Description
There was an issue before with the timer never cancelling, this was due to a new timer being created and replacing the reference to the old timer, as a result, it became impossible to actually cancel it.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1035)


dent50cent:
This now includes a potential fix for the edge case, discovered the signals library which let me create a kind of shutdown process for the Django server.

PR #1043, opened 2019-05-21
feat: delete game api

mariaFernando:
Renamed the existing GameView to reflect the fact that it lists user details. Created a ViewSet for accessing and deleting games and the necessary changes in urls, permissions (preview user permission) and tests.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1043)


mariaFernando:
___
*[aimmo/urls.py, line 49 at r1](https://reviewable.io/reviews/ocadotechnology/aimmo/1043#-LfTX09F5Ta4R4NPA4px:-LfTlMdA0ykpzyJSuoet:b-opw3q8) ([raw file](https://github.com/ocadotechnology/aimmo/blob/c577adb6cdb515cdec32972f86ec500b34543d3a/aimmo/urls.py#L49)):*
<details><summary><i>Previously, dent50cent (Paris Goldman-Smith) wrote…</i></summary><blockquote>

I believe there's a router class that can do this bit for you? I found [this](https://www.django-rest-framework.org/api-guide/routers/) which seems to route the standard actions for you.
</blockquote></details>

Do you know if that has an advantage over this method?

mariaFernando:
___
*[aimmo/urls.py, line 49 at r1](https://reviewable.io/reviews/ocadotechnology/aimmo/1043#-LfTX09F5Ta4R4NPA4px:-LfUg02UBUJI08jlIsHi:b-896fix) ([raw file](https://github.com/ocadotechnology/aimmo/blob/c577adb6cdb515cdec32972f86ec500b34543d3a/aimmo/urls.py#L49)):*
<details><summary><i>Previously, dent50cent (Paris Goldman-Smith) wrote…</i></summary><blockquote>

the main advantage would be auto-routing. so if we added a second action, like list, it would automatically match that action to a GET request for example. you can also specify custom routes for things like custom actions using routers I believe. If it's not easy to implement it I wouldn't worry too much.
</blockquote></details>

Done.

mariaFernando:
___
*[aimmo/views.py, line 113 at r1](https://reviewable.io/reviews/ocadotechnology/aimmo/1043#-LfTjn-MAc_KdNnSFUDM:-LfUgqWUA4tuARg1JhoX:bit7mi0) ([raw file](https://github.com/ocadotechnology/aimmo/blob/c577adb6cdb515cdec32972f86ec500b34543d3a/aimmo/views.py#L113)):*

do we need to also check that they are a teacher or is this enough? I guess at least the delete should be a teacher thing
</blockquote></details>

Yep, that sounds good, since the students with access to the games are also preview users

mariaFernando:
___
*[aimmo/models.py, line 60 at r5](https://reviewable.io/reviews/ocadotechnology/aimmo/1043#-Lg1bdxn4Nulf0HMBMUj:-Lg7cqig1l_rSsKIuBrD:b-a28adg) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5c03213d89c7cef5740d9d9802e3a43066871173/aimmo/models.py#L60)):*

it does but you have changed the format of the JSON response, those settings are no longer nested under a settings object. Changing how that response is read might be trickier than usual so I would recommend keeping the nested settings object format :) 

(also a test should have failed to catch this)
</blockquote></details>

I had a test for this but only to test the format I had before (with the settings attribute inside the game model, which is obviously not that great an idea). I'm not sure where this settings object is used but it's probably not tested where it should be. I'm not entirely sure how to keep the settings JSONfield inside the serializer without having a "settings" attribute inside the game model. Serializer relations?

mariaFernando:
___
*[aimmo/tests/test_views.py, line 398 at r5](https://reviewable.io/reviews/ocadotechnology/aimmo/1043#-Lg7P6Lm3sBOmb2th0T8:-Lg7Z1CT2tXBcXpr9gQD:b-uyv0f0) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5c03213d89c7cef5740d9d9802e3a43066871173/aimmo/tests/test_views.py#L398)):*

this test was checking the format for that API, that fact that it probably failed was probably an indication that whatever was API was using this serializer would have been broken (in this case, one of the containers in the Kurono game)
</blockquote></details>

I added that test because settings wasn't being displayed correctly - but it depended on having a "settings" attribute in the Game model. Will add it back in if we keep the structure as it was

mariaFernando:
___
*[aimmo/models.py, line 60 at r5](https://reviewable.io/reviews/ocadotechnology/aimmo/1043#-Lg1bdxn4Nulf0HMBMUj:-Lg81fWC-UQXS8D2o1h6:b-61bqpq) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5c03213d89c7cef5740d9d9802e3a43066871173/aimmo/models.py#L60)):*

you can nest a serializer within another serializer. @dent50cent should be able to help with this as he had to do it for a previous task :)  
</blockquote></details>

Yes I've been asking him for help but he seems to have struggled with the same thing

mariaFernando:
___
*[aimmo/tests/test_views.py, line 398 at r5](https://reviewable.io/reviews/ocadotechnology/aimmo/1043#-Lg7P6Lm3sBOmb2th0T8:-Lg81m28CPs7iBadpyOu:b-mv632j) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5c03213d89c7cef5740d9d9802e3a43066871173/aimmo/tests/test_views.py#L398)):*

did it also depend on having the json response having a settings attribute in it? but if so, that's a valid failure
</blockquote></details>

Yes, the test itself is very straightforward but depending on the format

PR #1046, opened 2019-05-29
Edits CodeCov comments settings

faucomte97:
Creates `.codecov.yml` to set up CodeCov comments in PRs.

The file contains a few settings to customise the behaviour of CodeCov comments. In this case, the comment will show a graph, the difference in coverage before and after the latest commit, as well as a detailed list of files and whether their coverage has gone up or down (specified in the `layout` setting`).

The `behavior` setting is set to `new`, which means that a new comment will be published at each new commit, and the old ones will be deleted. It doesn't seem to be doing this right now, instead it seems to be editing the old comment.

The `require_changes` setting is set to `false`, meaning that a new comment will be published even if the code coverage hasn't changed since the previous commit.

The `require_base` setting is set to `no` and the `require_head` is set to `yes` meaning that the comment will only be published if there is a coverage report on the head commit, but it doesn't require one on the base commit.

The `branches` setting is set to `!master, !development` meaning that the comments will be published on any branch other than the master branch and the development branch.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1046)


faucomte97:
Relaunching Travis to see if CodeCov plugin helps

faucomte97:
SUCCESS!

faucomte97:
Currently trying to figure out why CodeCov isn't posting any new comments after each commit

faucomte97:
However the report isn't right. It says 204+ files have been impacted when actually the coverage hasn't changed at all...

PR #1114, opened 2019-07-26
feat: added obstacles and babel parser

mariaFernando:
### Description
Added obstacles to the babylon project. Added onTerrainNode as obstacles' parent and passed game state from game component.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1114)


mariaFernando:
___
*[game_frontend/src/babylon/entities/obstacle.ts, line 41 at r12](https://reviewable.io/reviews/ocadotechnology/aimmo/1114#-Ll7Vij036iho8maFDFO:-LlB2Rg-87Z_qONlfggQ:b-x21ru1) ([raw file](https://github.com/ocadotechnology/aimmo/blob/cefb5d9d7c9ea73a4b8d5b31554e5eb749863e85/game_frontend/src/babylon/entities/obstacle.ts#L41)):*

I think you called it originalIndex (although id might be a better name perhaps?)
</blockquote></details>

My logic was, since the diff function is modular, it might receive as arguments arrays of objects that already have an id field. This way that wouldn't be an issue, I think

PR #1148, opened 2019-08-22
Fix games not stopping

dent50cent:
## Description
Some urls were still set to aimmo when they needed to say kurono, this prevented the games from stopping.

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1148)


dent50cent:
Not needed, changes were already present?

PR #1194, opened 2019-10-21
feat: center on avatar

mariaFernando:
### Description ###
On a game startup, the camera will centre on the current avatar and follow it around, then be released when the user drags on the map (panning).

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1194)


mariaFernando:
___
*[game_frontend/src/babylon/gameEngine.ts, line 34 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1194#-LrnPqsb92syDHjVemRh:-LrrzkZ701Bs8r2Fn06O:b-9yiieh) ([raw file](https://github.com/ocadotechnology/aimmo/blob/43ca1442b0266cb56de4acba9e28a97770abcff7/game_frontend/src/babylon/gameEngine.ts#L34)):*

let's see how far we can get on this, worst case scenario we can maybe uncenter and snap to the location is going to perhaps? (or keep the current logic)
</blockquote></details>

After talking to Florian and Laura, we've concluded that the snapping isn't a bit issue UX-wise - good to keep in mind and address later but not worth spending a lot of time on for now. Going to create a separate bug report for it

PR #1290, opened 2020-02-28
fix: Update styled-components, enzyme, jest-styled-components, material-ui and add core-js explicitly

mrniket:
This should resolve some of the warnings we get when we run the project

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1290)


mrniket:
closing due to inactivity

PR #1379, opened 2020-07-29
feat: Support different background colors for each era and different obstacle textures

mrniket:
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1379)


mrniket:
Background colours chosen found to be not good for accessibility, @lauracumming to provide updated background colours

Issue #1548, opened 2021-07-02
aimmo_setup.py needs fixes


dionizh:
Currently is not working. Would be good to fix it for linux and mac if possible.

dionizh:
@adamkirchberger Can you comment on this to indicate your interest to take it on? 🙂

adamkirchberger:
Hey, yes I'm happy to take this one 👍

PR #1568, opened 2021-08-25
fix: fix aimmo setup script for mac and linux

adamkirchberger:
fixes: #1548


dionizh (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1568)* status: 0 of 1 files reviewed, 13 unresolved discussions (waiting on @adamkirchberger)

___
*[aimmo_setup.py, line 19 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhwnsz492zBnn3Ymn1S:-Mhwnsz492zBnn3Ymn1T:b-5iy97w) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L19)):*
> ```Python
> MINIKUBE_VERSION = "1.21.0"
> KUBECTL_VERSION = "1.21.2"
> ```

Is it possible to get the latest major versions for these as they will get out of date soon? I'm already using 1.22.x 😆
___
*[aimmo_setup.py, line 39 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-MhwoZDf9YuXguZcmaS7:-MhwoZDf9YuXguZcmaS8:b-gim4bj) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L39)):*
> ```Python
>         "| just be kind if it doesn't work.                                                                         |\n"
>         "| You may be asked to enter your password during this setup.                                               |\n"
>         "+----------------------------------------------------------------------------------------------------------+\n"
> ```

Nice touch :D 

Can you add that it supports Mac and Linux and it does not currently work for Windows?
___
*[aimmo_setup.py, line 156 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhwu2fc9LHoa8t8pa4B:-Mhwu2fc9LHoa8t8pa4C:b4p5nnz) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L156)):*
> ```Python
>         minikube_start_profile,
>         helm_install_aimmo,
>     ]
> ```

I noticed there's no `pyenv` task here. Following the common setup here https://docs.codeforlife.education/git/common-setup there should be. If we assume that people have followed the common setup first before setting up aimmo, we can probably skip the pipenv. If not, I think it needs to do it from pyenv, then pipenv. What do you think? 
___
*[aimmo_setup.py, line 205 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-MhwsGx86RNxG-34eRXI:-MhwsGx86RNxG-34eRXJ:bbu44f) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L205)):*
> ```Python
> def _create_sudo_timestamp():
>     """
>     Request sudo access to create timestamp file for duration of setup
> ```

What's timestamp file? Not too clear what this is for 🤔
___
*[aimmo_setup.py, line 279 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhwv2Z68excUxmBLILM:-Mhwv2Z68excUxmBLILN:b-2d0zys) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L279)):*
> ```Python
> 
> 
> def yarn_add_parcel_bundler(os_type, arch_type):
> ```

This seems out of date. We can remove this. See: https://docs.codeforlife.education/git/aimmo-setup/frontend
___
*[aimmo_setup.py, line 286 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-MhwvBNG9cEvan2IJZBS:-MhwvBNG9cEvan2IJZBT:bc8lssk) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L286)):*
> ```Python
> def set_up_frontend_dependencies(os_type, arch_type):
>     if os_type in [OSType.MAC, OSType.LINUX]:
>         _cmd("cd ./game_frontend && sudo yarn")
> ```

Do we need sudo here?
___
*[aimmo_setup.py, line 302 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhwvgb_78pi4cprlG6m:-Mhwvgb_78pi4cprlG6n:bikjyk7) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L302)):*
> ```Python
>         _cmd("pip install pipenv")
>     elif os_type == OSType.WINDOWS:
>         pass
> ```

raise NotImplemented?
___
*[aimmo_setup.py, line 304 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhwvl662hqOO9F7pKZb:-Mhwvl662hqOO9F7pKZc:b-evepl0) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L304)):*
> ```Python
>         pass
>     else:
>         raise Exception
> ```

Probably does not need this. Or it could use a more descriptive error (e.g. OS not supported).
___
*[aimmo_setup.py, line 307 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhww2Ui88NEF7_LcBBn:-Mhww2Ui88NEF7_LcBBo:b-1yw1ce) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L307)):*
> ```Python
> 
> 
> def run_pipenv_install(os_type, arch_type):
> ```

Can this be renamed to `build_pipenv_virtualenv` as this sounds like it's installing pipenv.
___
*[aimmo_setup.py, line 441 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-MhwtYfIACluIJdIgGJ3:-MhwtYfIACluIJdIgGJ4:blu6aiu) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L441)):*
> ```Python
> def install_snap(os_type, arch_type):
>     if os_type == OSType.LINUX:
>         _cmd("sudo apt install snapd")
> ```

Can this be removed? There's already a snap install when installing kubectl.
___
*[aimmo_setup.py, line 465 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-MhwqYQeER2QEYH2s3n-:-MhwqYQeER2QEYH2s3n0:b-wsrnli) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L465)):*
> ```Python
>     the cmdtest package is installed, if it is we ask the user if we can remove it, if yes
>     we remove the package, if not the process continues without removing it.
>     """
> ```

Does cmdtest clash with yarn?
___
*[aimmo_setup.py, line 491 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhwqyif9TQhL3mQHRY5:-Mhwqyif9TQhL3mQHRY6:bww13tz) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L491)):*
> ```Python
> 
> 
> def add_aimmo_to_hosts_file(os_type, arch_type):
> ```

Mm I don't recognise this step. Is this meant to stay, or is this a test?
___
*[aimmo_setup.py, line 506 at r3](https://reviewable.io/reviews/ocadotechnology/aimmo/1568#-Mhwr92n9lWqRc4bmBKz:-Mhwr92n9lWqRc4bmBL-:bq92kal) ([raw file](https://github.com/ocadotechnology/aimmo/blob/5e6147df37189f40a01b2b479c6f9625f5aa637e/aimmo_setup.py#L506)):*
> ```Python
> 
> def configure_yarn_repo(os_type, arch_type):
>     comment = "configure_yarn"
> ```

This seems to be linux only? Do we need to check the OS type?

adamkirchberger:
Hey @dionizh, thanks for looking into this.

I've pushed some additional commits to address the points you mentioned.

*  The Minikube and Kubectl tasks now start by first getting the latest upstream version number, then checking if the local installed version matches and upgrading it if not. (or installing it if there is no version at all)
* The banner has been updated to include a message that the setup only supports Mac and Linux.
* Regarding Pyenv, I think blindly installing pyenv and changing a users local python version can be quite destructive to their system, in my opinion we should only force the python version used for aimmo not all the users tools and projects.
* Regarding sudo timestamp: when you run a command with sudo it asks you for the password once and then after 5 minutes of inactivity will re-ask you for the password. That is handled by a sudo timestamp file, so we're asking for a password so we can 1. start the 5 minute countdown and 2. avoid reaching half way during the setup to find the user doesn't have sudo privileges.
* Regarding the yarn and cmdtest tasks, I'm not really sure about these I want to remind you that I have taken the original aimmo_setup.py script and refactored it slightly so 99% of the tasks are coming from what was originally there for which I don't have context on. So if these tasks aren't required then we should absolutely remove them. 👍 
* Regarding snap, I've removed the snap install steps as kubectl can just be installed using the same steps used for Mac. In my opinion this is a better approach so the setup is consistent across platforms and to avoid adding another tool (snap) that may need to be troubleshooted.

Please let me know if you notice anything else.
Thanks 🙂

dionizh:
@adamkirchberger Thanks for the updates! To follow our process - can you please respond to the comments on Reviewable? You can simply say 'Done' if done, or reply to the comment if it needs further discussion. Each issue on Reviewable needs to be marked resolved before the PR can move forward, and it's currently waiting for your responses 🙂

PR #1607, opened 2022-01-18
fix: one class can have multiple games

dionizh:
more changes on portal coming

---
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1607)


dionizh (commented):
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1607)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dionizh)

___
*[aimmo/models.py, line 41 at r1](https://reviewable.io/reviews/ocadotechnology/aimmo/1607#-Mthm2M7EpNdTC__3WFx:-MthxnhK5ggWsAUxhkwo:b-sgebu7) ([raw file](https://github.com/ocadotechnology/aimmo/blob/8cb73bffabbea93128f8caa07cd4e03094cfbfcc/aimmo/models.py#L41)):*
<details><summary><i>Previously, razvan-pro (Razvan Mahu) wrote…</i></summary><blockquote>

Note that deleting a class will delete all games I think, even the archived ones - not sure if we want to do anything with this yet, just keep in mind and it will fix itself when we won't delete classes anymore (and archive instead, if we don't already do this :thinking:)
</blockquote></details>

Ah yea, for a moment I thought we've stopped deleting classes too, but we haven't. I'd say this should set to NULL then. And eventually we will archive classes too.

razvan-pro (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1607)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dionizh)

PR #1797, opened 2023-07-27
feat: Implement max games limit

faucomte97:
This change is [Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1797)


SKairinos (changes requested):
> Reviewed 14 of 18 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1797)* status: 15 of 18 files reviewed, 10 unresolved discussions (waiting on @faucomte97)

___
*[`aimmo/models.py` line 22 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nak_qfdDLnyNCxqfkMC:-Nak_qfdDLnyNCxqfkMD:b-akxzb6) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L22)):*
> ```Python
> ]
> 
> MAX_GAMES_LIMIT = 15
> ```

Let's add this as a setting.
from django.conf import settings
settings.MAX_GAMES_LIMIT



_Code quote:_
```Python
MAX_GAMES_LIMIT = 15
```
___
*[`aimmo/models.py` line 139 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakaRZt0hb1i3uZS3qR:-NakaRZt0hb1i3uZS3qS:b-66kyo3) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L139)):*
> ```Python
>         }
> 
>     def create_avatar_for_user(self, user):
> ```

let's move this logic to the Avatar.objects.create method. Creation of a model should be the responsibility of that model's object manager. This you would simply say:
avatar = Avatar.objects.create(owner=user, game=game)

Furthemore, let's add the generate_auth_token helper as a default callable of the auth_token field. It's safer since it should be impossible to create an avatar without an auth token. It would like something like:
def generate_game_auth_token():
       return generate_auth_token(16, 24)
auth_token = models.CharField(max_length=24, blank=True, default=generate_game_auth_token)
see: https://docs.djangoproject.com/en/4.2/ref/models/fields/#default



_Code quote:_
```Python
create_avatar_for_user
```
___
*[`aimmo/models.py` line 152 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nakd6J1DRwK4rdgAAdi:-Nakd6J1DRwK4rdgAAdj:b54t7tk) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L152)):*
> ```Python
> 
>     def save(self, **kwargs):
>         new_game = self.id is None
> ```

no need for variable, can write inline since it's only referenced once



_Code quote:_
```Python
new_game
```
___
*[`aimmo/models.py` line 154 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nakh4hx4qjraVWVVirV:-Nakh4hx4qjraVWVVirW:b5hkp0p) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L154)):*
> ```Python
>         new_game = self.id is None
> 
>         if new_game:
> ```

relocate everything inside this if block in the Game.objects.create



_Code quote:_
```Python
if new_game:
```
___
*[`aimmo/models.py` line 155 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakdOYb0bKh22SZEkto:-NakdOYb0bKh22SZEktp:b-89vxii) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L155)):*
> ```Python
> 
>         if new_game:
>             self.auth_token = generate_auth_token(32, 48)
> ```

see above comment about adding this as a default callable.



_Code quote:_
```Python
self.auth_token = generate_auth_token(32, 48)
```
___
*[`aimmo/models.py` line 156 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakdYAe9XCDwQdjEXqI:-NakdYAfEq6310Fz2ZD4:b-y892pw) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L156)):*
> ```Python
>         if new_game:
>             self.auth_token = generate_auth_token(32, 48)
>             self.generator = "Main"
> ```

set as default value. only set it in the create method if we want to override the default based on some condition(s). but since we always want it to be "Main", it can just be a default.



_Code quote:_
```Python
            self.generator = "Main"
```
___
*[`aimmo/models.py` line 179 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakhoQ6BggyHmbictPN:-NakhoQ6BggyHmbictPO:b-yig4a0) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L179)):*
> ```Python
>         def update(self, **kwargs) -> int:
>             if (
>                 kwargs.get("status", Game.STOPPED) == Game.RUNNING
> ```

no need for a default value. none is an acceptable return value



_Code quote:_
```Python
Game.STOPPED
```
___
*[`aimmo/models.py` line 180 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakipPW3bN_G11k4BuG:-NakipPW3bN_G11k4BuH:bvw9wtv) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L180)):*
> ```Python
>             if (
>                 kwargs.get("status", Game.STOPPED) == Game.RUNNING
>                 and Game.objects.filter(status=Game.RUNNING).count() + self.count() > MAX_GAMES_LIMIT
> ```

rethinking this, this doesn't seem like the correct check because we're not accounting for the fact there may be duplicate values between queryset 1 and 2. Therefore, we should first get the union of the 2 queryset to count only the distinct values.
see: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#union
we should create a test case to confirm



_Code quote:_
```Python
 Game.objects.filter(status=Game.RUNNING).count() + self.count()
```
___
*[`aimmo/models.py` line 190 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nakjldd-yZkW8mq7qRv:-Nakjldd-yZkW8mq7qRw:b-i2b1tz) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L190)):*
> ```Python
> @receiver(models.signals.pre_save, sender=Game)
> def check_game_limit(sender: t.Type[Game], instance: Game, **kwargs):
>     if instance.status == Game.RUNNING and sender.objects.filter(status=Game.RUNNING).count() >= MAX_GAMES_LIMIT:
> ```

to make the query more complete, let's also exclude the current instance from the queryset if the instance already exists (id is not None).
if instance.status == Game.RUNNING:
queryset = sender.objects.filter(status=Game.RUNNING)
if instance.id is not None:
queryset = queryset.exclude(id=instance.id)
if queryset.count() >= ....



_Code quote:_
```Python
 if instance.status == Game.RUNNING and sender.objects.filter(status=Game.RUNNING).count() 
```
___
*[`aimmo/views.py` line 51 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakkjCR6hVh54hg_X2b:-NakkjCR6hVh54hg_X2c:bb1vv60) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/views.py#L51)):*
> ```Python
>         avatar = game.avatar_set.get(owner=request.user)
>     except Avatar.DoesNotExist:
>         avatar = game.create_avatar_for_user(request.user)
> ```

Avatar.objetcs.create(game=game, owner=request.user)



_Code quote:_
```Python
 game.create_avatar_for_user(request.user)
```

faucomte97 (commented):
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1797)* status: 15 of 18 files reviewed, 10 unresolved discussions (waiting on @faucomte97 and @SKairinos)

___
*[`aimmo/models.py` line 22 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nak_qfdDLnyNCxqfkMC:-Naso4U358usnYJqKyjZ:b-ziab7i) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L22)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

Let's add this as a setting.
from django.conf import settings
settings.MAX_GAMES_LIMIT
</blockquote></details>

Leaving it in here for now cos moving it to settings causes some problems with the mocking in the tests... Can revisit later
___
*[`aimmo/models.py` line 139 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakaRZt0hb1i3uZS3qR:-NaskIa_3UUMxcT-OFt-:b-5hkzz4) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L139)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

let's move this logic to the Avatar.objects.create method. Creation of a model should be the responsibility of that model's object manager. This you would simply say:
avatar = Avatar.objects.create(owner=user, game=game)

Furthemore, let's add the generate_auth_token helper as a default callable of the auth_token field. It's safer since it should be impossible to create an avatar without an auth token. It would like something like:
def generate_game_auth_token():
       return generate_auth_token(16, 24)
auth_token = models.CharField(max_length=24, blank=True, default=generate_game_auth_token)
see: https://docs.djangoproject.com/en/4.2/ref/models/fields/#default
</blockquote></details>

Left logic in save() in the end.
___
*[`aimmo/models.py` line 152 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nakd6J1DRwK4rdgAAdi:-NaqI08R-8KsLOI_RSkC:b-896fix) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L152)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

no need for variable, can write inline since it's only referenced once
</blockquote></details>

Done.
___
*[`aimmo/models.py` line 154 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nakh4hx4qjraVWVVirV:-NaqI0zy8DB9ByfTEebv:b-896fix) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L154)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

relocate everything inside this if block in the Game.objects.create
</blockquote></details>

Done.
___
*[`aimmo/models.py` line 155 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakdOYb0bKh22SZEkto:-NaqKJ_57DdaccixJZ2f:b-896fix) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L155)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

see above comment about adding this as a default callable.
</blockquote></details>

Done.
___
*[`aimmo/models.py` line 156 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakdYAe9XCDwQdjEXqI:-NaqKK155dfQlRqYJDfS:broc2t1) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L156)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

set as default value. only set it in the create method if we want to override the default based on some condition(s). but since we always want it to be "Main", it can just be a default.
</blockquote></details>

Already a default
___
*[`aimmo/models.py` line 179 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakhoQ6BggyHmbictPN:-NaqKNOo6VGkKD7whbu_:b-896fix) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L179)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

no need for a default value. none is an acceptable return value
</blockquote></details>

Done.
___
*[`aimmo/models.py` line 180 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakipPW3bN_G11k4BuG:-NaqMN1GDt19z3b2IorQ:b-896fix) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L180)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

rethinking this, this doesn't seem like the correct check because we're not accounting for the fact there may be duplicate values between queryset 1 and 2. Therefore, we should first get the union of the 2 queryset to count only the distinct values.
see: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#union
we should create a test case to confirm
</blockquote></details>

Done.
___
*[`aimmo/models.py` line 190 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nakjldd-yZkW8mq7qRv:-NaqM_hI22MaWCFlbZB0:b-1e5yhr) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L190)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

to make the query more complete, let's also exclude the current instance from the queryset if the instance already exists (id is not None).
if instance.status == Game.RUNNING:
queryset = sender.objects.filter(status=Game.RUNNING)
if instance.id is not None:
queryset = queryset.exclude(id=instance.id)
if queryset.count() >= ....
</blockquote></details>

I don't follow what you're saying here - why would we want to do that?
___
*[`aimmo/views.py` line 51 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakkjCR6hVh54hg_X2b:-NaqxGr478g-6Y539V7S:b-896fix) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/views.py#L51)):*
<details><summary><i>Previously, SKairinos (Stefan Kairinos) wrote…</i></summary><blockquote>

Avatar.objetcs.create(game=game, owner=request.user)
</blockquote></details>

Done.

SKairinos (changes requested):
> Reviewed 1 of 18 files at r1, 5 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1797)* status: 20 of 21 files reviewed, 2 unresolved discussions (waiting on @faucomte97)

___
*[`aimmo/models.py` line 155 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-NakdOYb0bKh22SZEkto:-NbE9_yfCE1X8aJg1Hsh:b8pquf4) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L155)):*
<details><summary><i>Previously, faucomte97 (Florian Aucomte) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

default missing on model field
___
*[`aimmo/models.py` line 190 at r2](https://reviewable.io/reviews/ocadotechnology/aimmo/1797#-Nakjldd-yZkW8mq7qRv:-NbEBy7m0oCyhrXmjoB4:b-rb0m67) ([raw file](https://github.com/ocadotechnology/aimmo/blob/35ba7a25ed6d3033a6313791f91ad30f65f5e14f/aimmo/models.py#L190)):*
<details><summary><i>Previously, faucomte97 (Florian Aucomte) wrote…</i></summary><blockquote>

I don't follow what you're saying here - why would we want to do that?
</blockquote></details>

Basically, I'm saying that instance MAY ALREADY BE RUNNNING and included in the queryset. Therefore, the instance should be excluded from the queryset and not raise an exception. This scenario could occur when updating another field on a game instance that is already running.

SKairinos (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1797)* status: all files reviewed, 1 unresolved discussion (waiting on @faucomte97)

SKairinos (approved):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1797)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @faucomte97)

faucomte97 (commented):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ocadotechnology/aimmo/1797)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @faucomte97)

